PR #6, opened 2017-10-21
[WIP] Readline prompt in menu and ask y/n

ailisp:
ISSUE: When choose item in menu and answer yes/no in appliance console, user can't use arrow key to navigate and delete characters. Answers of long strings have been fixed before (ip address, enter a path, etc.), this PR is for menu and ask y/n.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1446249

\cc @yrudman @carbonin


ailisp:
Note: the ask menu part is still too tricky to make it looks the same as previous no readline prompt. This part is still work in progress

kbrock:
when you are fixing the specs, would you consider squashing these commits?
Feel it would be easier to review in the future

ailisp:
@kbrock Yes, will squash that. Thanks. I found now there is just too many hacks to get around highline's bug in readline mode (done), and also pass specs (wip, 2/3 finished).

ailisp:
@carbonin Finally got all menu, ask y/n and are_you_sure can also use arrow keys as you said in summer and Luke said in his recent comment. Because I use much hacks to make all the things work, can you have a review for all these stuff before I squash them to neat comments? This fix is very ugly, but almost the only way I can find. I think you must have better ways to do some of these `ask*` methods, but some of them I've tried simply not working. I can tell you the detail about that when we're both in office.
I also considered a new version of highline gem may fix these bugs. But as their most recent 1.8.7 version and 2.0.0-preview didn't fix this inconsistency issue. So it's faster to fix in our side. Thank you for review!

ailisp:
Closed since it's fixed upstream highline gem: https://github.com/JEG2/highline/pull/222#event-1354605151. Will make another PR when highline release a new stable version.

Issue #10, opened 2017-10-25
DDBB appliance as GPT Partition Table


jparrill:
Hi there,

When you try to configure database on a HDD/SSD with a size more than 2 TB, the Appliance_console gives an error:

> Initialize postgresql disk starting
> Initialize postgresql disk failed with error - parted -s /dev/vdf mkpart primary 0% 100% exit code: 1.

In the emv.log you can see:
> appliance_console.log:E, [2017-10-24T13:56:37.146463 #26751] ERROR -- : MIQ(ApplianceConsole::InternalDatabaseConfiguration#initialize_postgresql_disk)  Command failed: parted -s /dev/vde mkpart primary 0% 100% exit code: 1. Error: Error: partition length of 6442448896 sectors exceeds the msdos-partition-table-imposed maximum of 4294967295

The problem here is the appliance console component, by default are defined as msdos part table, we need to set as GPT.

Fryguy:
cc @carbonin

bdunne:
Fixed by #23

PR #11, opened 2017-10-25
#10 - Add GPT as partition table when disk size are bigger than 2 TB

jparrill:
### Implemented Solution
Integrated with LinuxAdmin library, passing from **logical_volume_management.rb** as parameter the partition_type to **create_partition_table** function.

### Solves Issue: #10

### DIFF
--- a/lib/manageiq/appliance_console/logical_volume_management.rb
+++ b/lib/manageiq/appliance_console/logical_volume_management.rb
@@ -41,10 +41,19 @@ module ApplianceConsole
     private
 
     def create_partition_to_fill_disk
-      disk.create_partition_table
-      AwesomeSpawn.run!("parted -s #{disk.path} mkpart primary 0% 100%")
-
+      # Check if you need to create a GPT part table or a MSDOS one in base of
+      # max size of partition table
+      max_msdos_ptable_size = 2199023255552
       self.disk = LinuxAdmin::Disk.local.find { |d| d.path == disk.path }
+
+      if self.disk.size  >= max_msdos_ptable_size
+        partition_type = 'gpt'
+      else
+        partition_type = 'msdos'
+      end
+      disk.create_partition_table(partition_type)
+
+      AwesomeSpawn.run!("parted -s #{disk.path} mkpart primary 0% 100%")
       @partition = disk.partitions.first
     end
 
--- a/spec/logical_volume_management_spec.rb
+++ b/spec/logical_volume_management_spec.rb
@@ -29,7 +29,8 @@ describe ManageIQ::ApplianceConsole::LogicalVolumeManagement do
 
   describe "#setup" do
     before do
-      expect(@disk_double).to receive(:create_partition_table)
+      expect(@disk_double).to receive(:create_partition_table).with("gpt")
+      expect(@disk_double).to receive(:create_partition_table).with("msdos")
       expect(@disk_double).to receive(:partitions).and_return([:fake_partition])
       @config.disk = @disk_double
 
--- a/bin/appliance_console
+++ b/bin/appliance_console
@@ -173,7 +173,7 @@ To modify the configuration, use a web browser to access the management page.
           new_mask = ask_for_ipv4("Netmask", mask)
           new_gw   = ask_for_ipv4("Gateway", gw)
           new_dns1 = ask_for_ipv4("Primary DNS", dns1)
-          new_dns2 = ask_for_ipv4_or_none("Secondary DNS (Enter 'none' for no value)")
+          new_dns2 = ask_for_ipv4_or_none("Secondary DNS (Enter 'none' for no value)", dns2)
 
           new_search_order = ask_for_many("domain", "Domain search order", order)
 
@@ -225,7 +225,7 @@ Static Network Configuration
           new_prefix = ask_for_integer('IPv6 prefix length', 1..127, eth0.prefix6 || 64)
           new_gw = ask_for_ipv6('Default gateway', eth0.gateway6)
           new_dns1 = ask_for_ip('Primary DNS', dns1)
-          new_dns2 = ask_for_ipv6_or_none("Secondary DNS (Enter 'none' for no value)")
+          new_dns2 = ask_for_ipv6_or_none("Secondary DNS (Enter 'none' for no value)", dns2)
 
           new_search_order = ask_for_many('domain', 'Domain search order', order)
 
--- a/lib/manageiq-appliance_console.rb
+++ b/lib/manageiq-appliance_console.rb
@@ -24,6 +24,7 @@ require 'manageiq/appliance_console/logger'
 require 'manageiq/appliance_console/logging'
 
 require 'manageiq-gems-pending'
+require 'highline'
 
 require 'manageiq/appliance_console/certificate'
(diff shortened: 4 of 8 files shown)


lib/manageiq/appliance_console/logical_volume_management.rb
```diff
@@ -41,10 +41,19 @@ def setup
     private
 
     def create_partition_to_fill_disk
-      disk.create_partition_table
-      AwesomeSpawn.run!("parted -s #{disk.path} mkpart primary 0% 100%")
-
+      # Check if you need to create a GPT part table or a MSDOS one in base of
+      # max size of partition table
+      max_msdos_ptable_size = 2199023255552
```
carbonin: This feels like it should be a constant in `LinuxAdmin::Disk` rather than a variable here.

lib/manageiq/appliance_console/logical_volume_management.rb
```diff
@@ -41,10 +41,19 @@ def setup
...
       self.disk = LinuxAdmin::Disk.local.find { |d| d.path == disk.path }
+
+      if self.disk.size  >= max_msdos_ptable_size
```
carbonin: Extra space not needed.

lib/manageiq/appliance_console/logical_volume_management.rb
```diff
@@ -41,10 +41,19 @@ def setup
...
+        partition_type = 'gpt'
+      else
+        partition_type = 'msdos'
+      end
+      disk.create_partition_table(partition_type)
+
+      AwesomeSpawn.run!("parted -s #{disk.path} mkpart primary 0% 100%")
```
carbonin: I think we should probably use `LinuxAdmin::Disk#create_partition` here, but probably not in this patch.

spec/logical_volume_management_spec.rb
```diff
@@ -29,7 +29,8 @@
 
   describe "#setup" do
     before do
-      expect(@disk_double).to receive(:create_partition_table)
+      expect(@disk_double).to receive(:create_partition_table).with("gpt")
+      expect(@disk_double).to receive(:create_partition_table).with("msdos")
```
NickLaMuro: @padajuan Noticed you asked about fixing the tests a week or so ago, and I have finally taken some time to give you an answer.

Without going into too much detail, what you need to do here is basically first make sure that you are stubbing the `#size` method on the `@disk_double`, since that wasn't expected to be be hit prior to your patch in the tests.  That said, you also want to make sure you are returning a size that would match what the existing tests were expecting previously, so something below the `2199023255552` you have in your patch.  The stub code should looks something like this:

```ruby
expect(@disk_double).to receive(:size).with(2000000000000)
```

The other thing is now that you have an variable argument for the `#create_partition_table` method, which you have started to stub, but unfortunately used a bit incorrectly.  What you have here will expect that method to be called once with the `"gpt"` arg and once with the `"msdos"` in **EACH** of the spec blocks below (the `it` blocks), instead of what you were probably intending, which was to allow it to accept either of the arguments.

I am attaching a diff on how you could go about this, as well as adding another test for the 2TB case.  I didn't get fancy on making this DRY-er, and there is a decent amount of copy-pasta involved, but it gets the job done.


```diff
diff --git a/spec/logical_volume_management_spec.rb b/spec/logical_volume_management_spec.rb
index 621ecda..077a97d 100644
--- a/spec/logical_volume_management_spec.rb
+++ b/spec/logical_volume_management_spec.rb
@@ -29,8 +29,6 @@ describe ManageIQ::ApplianceConsole::LogicalVolumeManagement do
 
   describe "#setup" do
     before do
-      expect(@disk_double).to receive(:create_partition_table).with("gpt")
-      expect(@disk_double).to receive(:create_partition_table).with("msdos")
       expect(@disk_double).to receive(:partitions).and_return([:fake_partition])
       @config.disk = @disk_double
 
@@ -55,6 +53,8 @@ describe ManageIQ::ApplianceConsole::LogicalVolumeManagement do
     end
 
     it "sets up the logical disk when mount point is not a symbolic link" do
+      expect(@disk_double).to receive(:size).and_return(2000000000000)
+      expect(@disk_double).to receive(:create_partition_table).with("msdos")
       @tmp_mount_point = @config.mount_point = Pathname.new(Dir.mktmpdir)
       expect(@fstab).to receive(:write!)
       expect(FileUtils).to_not receive(:mkdir_p).with(@config.mount_point)
@@ -71,6 +71,8 @@ describe ManageIQ::ApplianceConsole::LogicalVolumeManagement do
     end
 
     it "recreates the new mount point and sets up the logical disk when mount point is a symbolic link" do
+      expect(@disk_double).to receive(:size).and_return(2000000000000)
+      expect(@disk_double).to receive(:create_partition_table).with("msdos")
       @tmp_mount_point = Pathname.new(Dir.mktmpdir)
       @config.mount_point = Pathname.new("#{Dir.tmpdir}/#{@spec_name}")
       FileUtils.ln_s(@tmp_mount_point, @config.mount_point)
@@ -91,6 +93,8 @@ describe ManageIQ::ApplianceConsole::LogicalVolumeManagement do
     end
 
     it "skips update if mount point is in fstab when mount point is in already fstab" do
+      expect(@disk_double).to receive(:size).and_return(2000000000000)
+      expect(@disk_double).to receive(:create_partition_table).with("msdos")
       @tmp_mount_point = @config.mount_point = Pathname.new(Dir.mktmpdir)
       expect(FileUtils).to_not receive(:mkdir_p).with(@config.mount_point)
       expect(AwesomeSpawn).to receive(:run!)
@@ -108,5 +112,23 @@ describe ManageIQ::ApplianceConsole::LogicalVolumeManagement do
       expect(@config.logical_volume).to eq(@fake_logical_volume)
       expect(@fstab.entries.count).to eq(1)
     end
+
+    it "uses gpt partition table when disk size is over 2TB" do
+      expect(@disk_double).to receive(:size).and_return(2199023255552)
+      expect(@disk_double).to receive(:create_partition_table).with("gpt")
+      @tmp_mount_point = @config.mount_point = Pathname.new(Dir.mktmpdir)
+      expect(@fstab).to receive(:write!)
+      expect(FileUtils).to_not receive(:mkdir_p).with(@config.mount_point)
+      expect(AwesomeSpawn).to receive(:run!)
+        .with("mount",
+              :params => {"-t" => @config.filesystem_type, nil => ["/dev/vg_test/lv_test", @config.mount_point]})
+
+      @config.setup
+      expect(@config.partition).to eq(:fake_partition)
+      expect(@config.physical_volume).to eq(:fake_physical_volume)
+      expect(@config.volume_group).to eq(:fake_volume_group)
+      expect(@config.logical_volume).to eq(@fake_logical_volume)
+      expect(@fstab.entries.count).to eq(1)
+    end
   end
 end
```

Hope this helps.

lib/manageiq/appliance_console/logical_volume_management.rb
Fryguy: also prefer `2.terabytes` if active_support is available

ailisp: @carbonin currently not in LinuxAdmin::Disk, should I add it. `2.terabytes` works fine though

bdunne:
Fixed in #23

PR #12, opened 2017-10-25
Keep previous second dns settings when reconfig network

ailisp:
ISSUE: appliance_console doesn't keep previous second dns settings when reconfigure network, which should be kept as default value for new second dns.
BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1439348

\cc @yrudman @gtanzillo


yrudman:
looks good 👍

PR #13, opened 2017-10-25
recover from crash when setting appliance as HA primary node and no region exist

ailisp:
ISSUE:
When setting database replication as HA primary code, and there is no exist region, appliance console will crash

BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1500447

\cc @yrudman @gtanzillo

### DIFF
--- a/lib/manageiq/appliance_console/database_replication.rb
+++ b/lib/manageiq/appliance_console/database_replication.rb
@@ -71,7 +71,7 @@ Replication Server Configuration
         primary_region_number =
           pg_conn.exec("SELECT last_value FROM miq_databases_id_seq").first["last_value"].to_i / 1_000_000_000_000
         self.cluster_name = "miq_region_#{primary_region_number}_cluster"
-      rescue PG::ConnectionBad => e
+      rescue PG::ConnectionBad, PG::UndefinedTable => e
         say("Failed to get primary region number #{e.message}")
         logger.error("Failed to get primary region number #{e.message}")
         return false


yrudman:
👍  looks good to me

lib/manageiq/appliance_console/database_replication.rb
```diff
@@ -75,6 +75,10 @@ def generate_cluster_name
         say("Failed to get primary region number #{e.message}")
         logger.error("Failed to get primary region number #{e.message}")
         return false
+      rescue PG::UndefinedTable
```
bdunne: Any reason not to rescue this with the `PG::ConnectionBad` error above?  Is there something in the message that you don't want to display?

yrudman: I think it looks more friendly to user to communicate error in known terms (`no region exists`) than in some internal terms (`relation "miq_databases_id_seq" does not exist`)

ailisp: @bdunne In this particular BZ, appliance_console will raise `PG::UndefinedTable` instead of `PG::ConnectionBad` because ap tries to `SELECT last_value FROM miq_databases_id_seq` but `miq_databases_id_seq` doesn't exist (because of improper settings, regions not created). So here I rescue also for this error

bdunne: The `rescue` on https://github.com/ManageIQ/manageiq-appliance_console/pull/13/files#diff-96034652e522899e72c0060781012f58L74 seems to be that we can't even connect to the database.  This one is that we can connect but the database hasn't been regioned yet.  Is that correct?  Do the two situations have different resolutions?

ailisp:
@bdunne sorry, just got your idea. Wasn't think of I can put two exceptions in one rescue. Thank you

PR #14, opened 2017-11-03
require highline to get tests passing

bdunne:
It's already required in `bin/appliance_console`, but I don't like the idea of adding the require to the spec file.


Fryguy:
Since we do "real" releases here we shouldn't use gaprindashvili labels.  I'll change the release script to deal with that.

PR #15, opened 2017-11-03
[WIP] use GPT partition table to support partition > 2TB

ailisp:
ISSUE:
Can't create partition > 2 TB becuase using MBR partition table doesn't support, and that lead to initialize postgresql database failed. Use GPT partition table to fix that.
BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1506997

\cc @gtanzillo @yrudman @bdunne 
@miq-bot add-label wip, bug

@gtanzillo This should work but we'd better find a more than 2TB hard disk to test that before merge


ailisp:
Close because duplicate with https://github.com/ManageIQ/manageiq-appliance_console/pull/11/files

PR #16, opened 2017-11-03
correct error message for create internal database without unpartitioned disk available

ailisp:
Issue:
Now create internal database without unpartitioned disk, applaince console will ask are you sure for that, and whether choose yes or no will lead to configure fail. So we just show configure fail because of no unpartitioned disk and mount point is not a disk, and don't ask this not useful "are you sure"
BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1508958

\cc @yrudman @gtanzillo 
@miq-bot  add-label wip, bug

### DIFF
--- a/lib/manageiq/appliance_console/internal_database_configuration.rb
+++ b/lib/manageiq/appliance_console/internal_database_configuration.rb
@@ -65,11 +65,11 @@ module ApplianceConsole
     end
 
     def choose_disk
-      @disk = ask_for_disk("database disk")
+      @disk = ask_for_disk("database disk", false, true)
     end
 
     def check_disk_is_mount_point
-      error_message = "The disk for database must be a mount point"
+      error_message = "Internal databases require a volume mounted at #{mount_point}. Please add an unpartitioned disk and try again."
       raise error_message unless disk || pg_mount_point?
     end
 
--- a/lib/manageiq/appliance_console/prompts.rb
+++ b/lib/manageiq/appliance_console/prompts.rb
@@ -144,12 +144,12 @@ module ApplianceConsole
       just_ask(prompt, default, INT_REGEXP, "an integer", Integer) { |q| q.in = range if range }
     end
 
-    def ask_for_disk(disk_name, verify = true)
+    def ask_for_disk(disk_name, verify = true, silent = false)
       require "linux_admin"
       disks = LinuxAdmin::Disk.local.select { |d| d.partitions.empty? }
 
       if disks.empty?
-        say "No partition found for #{disk_name}. You probably want to add an unpartitioned disk and try again."
+        say("No partition found for #{disk_name}. You probably want to add an unpartitioned disk and try again.") unless silent
       else
         default_choice = disks.size == 1 ? "1" : nil
         disk = ask_with_menu(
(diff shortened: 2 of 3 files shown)


ailisp:
Seems need to wait after merge https://github.com/ManageIQ/manageiq-appliance_console/pull/14 to pass travis tests.
Local tests are fine

ailisp:
@miq-bot remove-label wip

ailisp:
@miq-bot remove-label wip

lib/manageiq/appliance_console/internal_database_configuration.rb
```diff
@@ -65,11 +65,11 @@ def ask_questions
     end
 
     def choose_disk
-      @disk = ask_for_disk("database disk")
+      @disk = ask_for_disk_silently("database disk")
     end
 
     def check_disk_is_mount_point
-      error_message = "The disk for database must be a mount point"
+      error_message = "The disk for database must be a mount point. Please add an unpartitioned disk and try again."
```
carbonin: While we're here, can we improve the first part of this message too? Maybe:

```ruby
"Internal databases require a volume mounted at #{mount_point}. Please add an unpartitioned disk and try again."
```

lib/manageiq/appliance_console/prompts.rb
```diff
@@ -168,6 +168,25 @@ def ask_for_disk(disk_name, verify = true)
       disk
     end
 
+    # Used for internal database configuration. Disk can be nil but in that case,
+    # our default pg mount point must already been manually mounted to a empty
+    # disk by user. Just return nil here and check in internal database config
+    def ask_for_disk_silently(disk_name)
```
carbonin: Couldn't we just add a `silent = false` parameter to `ask_for_disk` then change the caller that needs this to `ask_for_disk("database disk", false, true)`?

Then the silent parameter would govern whether we print the prompt on line 152. I think that does exactly what this method does.

lib/manageiq/appliance_console/prompts.rb
```diff
@@ -160,10 +162,11 @@ def ask_for_disk(disk_name, verify = true)
           q.choice("Don't partition the disk") { nil }
         end
       end
-
-      if verify && disk.nil?
-        say ""
-        raise MiqSignalError unless are_you_sure?(" you don't want to partition the #{disk_name}")
+      unless silent
```
carbonin: We already have a guard here (verify) I don't think we also need to add this conditional.

It wouldn't make very much sense to pass verify true with silent also true, would it?

lib/manageiq/appliance_console/prompts.rb
```diff
@@ -144,12 +144,14 @@ def ask_for_integer(prompt, range = nil, default = nil)
...
     end
 
-    def ask_for_disk(disk_name, verify = true)
+    def ask_for_disk(disk_name, verify = true, silent = false)
       require "linux_admin"
       disks = LinuxAdmin::Disk.local.select { |d| d.partitions.empty? }
 
       if disks.empty?
-        say "No partition found for #{disk_name}. You probably want to add an unpartitioned disk and try again."
+        unless silent
```
carbonin: This can be one line I think

lib/manageiq/appliance_console/prompts.rb
ailisp: It wouldn't. Thanks

yrudman:
also fixed this BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1513465

PR #17, opened 2017-11-11
Add database active/standby replication configuration

rsevilla87:
Configuring active/standby cluster of PostgreSQL in a non-interactive way should be available, implementing this feature in _appliance_console_cli_ tool will be useful to help users to automatize database cluster installation and configuration.

### DIFF
--- a/lib/manageiq/appliance_console/cli.rb
+++ b/lib/manageiq/appliance_console/cli.rb
@@ -79,6 +79,14 @@ module ApplianceConsole
       options[:db_hourly_maintenance]
     end
 
+    def set_replication?
+      options[:cluster_node_number] && options[:password] && replication_params?
+    end
+
+    def replication_params?
+      options[:replication] == "primary" || (options[:replication] == "standby" && options[:primary_host])
+    end
+
     def initialize(options = {})
       self.options = options
     end
@@ -116,6 +124,11 @@ module ApplianceConsole
         opt :force_key, "Forcefully create encryption key", :type => :boolean, :short => "f"
         opt :sshlogin,  "SSH login",         :type => :string,                 :default => "root"
         opt :sshpassword, "SSH password",    :type => :string
+        opt :replication, "Configure database replication as primary or standby", :type => :string, :short => :none
+        opt :primary_host, "Primary database host IP address", :type => :string, :short => :none
+        opt :standby_host, "Standby database host IP address", :type => :string, :short => :none
+        opt :auto_failover, "Configure Replication Manager (repmgrd) for automatic failover", :type => :bool, :short => :none
+        opt :cluster_node_number, "Database unique cluster node number", :type => :integer, :short => :none
         opt :verbose,  "Verbose",            :type => :boolean, :short => "v"
         opt :dbdisk,   "Database Disk Path", :type => :string
         opt :logdisk,  "Log Disk Path",      :type => :string
@@ -141,7 +154,8 @@ module ApplianceConsole
     def run
       Trollop.educate unless set_host? || key? || database? || tmp_disk? || log_disk? ||
                              uninstall_ipa? || install_ipa? || certs? || extauth_opts? ||
-                             time_zone? || set_server_state? || db_hourly_maintenance?
+                             time_zone? || set_server_state? || db_hourly_maintenance? ||
+                             set_replication?
       if set_host?
         system_hosts = LinuxAdmin::Hosts.new
         system_hosts.hostname = options[:host]
@@ -151,6 +165,7 @@ module ApplianceConsole
       end
       create_key if key?
       set_db if database?
+      set_replication if set_replication?
       set_time_zone if time_zone?
       config_db_hourly_maintenance if db_hourly_maintenance?
       config_tmp_disk if tmp_disk?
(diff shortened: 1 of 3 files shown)


mvazquezc:
+1

rcarrata:
+1

lib/manageiq/appliance_console/cli.rb
```diff
@@ -8,6 +8,8 @@ def say(arg)
   end
 end
 
+NETWORK_INTERFACE = "eth0".freeze
```
carbonin: This is a very strange thing to have to do, can you define this constant in `DatabaseReplicationStandby` rather than here?

lib/manageiq/appliance_console/cli.rb
```diff
@@ -232,6 +249,23 @@ def set_external_db
       config.post_activation
     end
 
+    def replication
```
carbonin: Can you make this method name something with a verb to match the others (eg. `set_db`, `create_key`, `config_tmp_disk`)? It reads a but strange as is. Maybe `configure_replication`?

lib/manageiq/appliance_console/cli.rb
```diff
@@ -116,6 +126,11 @@ def parse(args)
         opt :force_key, "Forcefully create encryption key", :type => :boolean, :short => "f"
         opt :sshlogin,  "SSH login",         :type => :string,                 :default => "root"
         opt :sshpassword, "SSH password",    :type => :string
+        opt :replication, "Configure database replication as primary or standby", :type => :string, :short => :none
+        opt :primary_host, "Primary database host IP address", :type => :string, :short => :none
+        opt :standby_host, "Standby database host IP address", :type => :string, :short => :none
+        opt :repmgrd, "Configure Replication Manager (repmgrd) for automatic failover", :type => :bool, :short => :none
```
carbonin: I think we can make this option name better, if people don't know anything about `repmgrd`, it won't be clear what this does.

I think `auto_failover` would be better.

lib/manageiq/appliance_console/cli.rb
```diff
@@ -232,6 +249,23 @@ def set_external_db
...
+      if options[:replication] == "primary"
+        db_replication = ManageIQ::ApplianceConsole::DatabaseReplicationPrimary.new
+        say("Configuring Server as Primary")
+      else
+        db_replication = ManageIQ::ApplianceConsole::DatabaseReplicationStandby.new
+        say("Configuring Server as Standby")
+        db_replication.disk = disk_from_string(options[:dbdisk])
+        db_replication.primary_host = options[:primary_host]
+        db_replication.standby_host = options[:standby_host]
+        db_replication.run_repmgrd_configuration = options[:repmgrd] ? "enabled" : "disabled"
```
carbonin: The `run_repmgrd_configuration` attribute should be `true` or `false`.

As is, this would always configure repmgrd as both strings are considered "truthy"

lib/manageiq/appliance_console/cli.rb
```diff
@@ -232,6 +249,23 @@ def set_external_db
...
+    def replication
+      if options[:replication] == "primary"
+        db_replication = ManageIQ::ApplianceConsole::DatabaseReplicationPrimary.new
+        say("Configuring Server as Primary")
+      else
+        db_replication = ManageIQ::ApplianceConsole::DatabaseReplicationStandby.new
+        say("Configuring Server as Standby")
+        db_replication.disk = disk_from_string(options[:dbdisk])
+        db_replication.primary_host = options[:primary_host]
+        db_replication.standby_host = options[:standby_host]
```
carbonin: This will default to the current appliance's address. So we can make this parameter optional and only set the attribute on `db_replication` if it was provided.

lib/manageiq/appliance_console/cli.rb
```diff
@@ -232,6 +249,23 @@ def set_external_db
...
+        db_replication.run_repmgrd_configuration = options[:repmgrd] ? "enabled" : "disabled"
+      end
+      db_replication.node_number = options[:cluster_node_number]
+      db_replication.database_password = options[:password]
```
carbonin: You also need to set `database_name` and `database_user` if they were provided. Both the classes set sane defaults (vmdb_production/root), but we should be able to change those using the cli.

lib/manageiq/appliance_console/cli.rb
rsevilla87: This variable is also defined in appliance_console.rb, that's why I've defined it here though I think  it should be removed in both files.

carbonin: I know, I think you should re-define it in `DatabaseReplication`. That way you don't need to define it here, where it is not used.

rsevilla87: Yep, makes sense, I'll work in this and all the other proposed changes. Thanks for the review

spec/cli_spec.rb
```diff
@@ -528,6 +528,53 @@
     end
   end
 
+  context "#set_replication" do
+    let(:replication_double) { double(:activate => true) }
```
carbonin: Rather than a double here, it would be better to use an actual instance of the class.

That way, rather than stubbing all the assignments, you will be able to assert on the actual value of the attributes. Then you just need to stub the `#activate` method.

lib/manageiq/appliance_console/cli.rb
```diff
@@ -232,6 +247,25 @@ def set_external_db
...
+    def set_replication
+      if options[:replication] == "primary"
+        db_replication = ManageIQ::ApplianceConsole::DatabaseReplicationPrimary.new
+        say("Configuring Server as Primary")
+      else
+        db_replication = ManageIQ::ApplianceConsole::DatabaseReplicationStandby.new
+        say("Configuring Server as Standby")
+        db_replication.disk = disk_from_string(options[:dbdisk])
+        db_replication.primary_host = options[:primary_host]
+        db_replication.standby_host = options[:standby_host] unless options[:standby_host].nil?
```
carbonin: These are all double negatives, can these be `if options[:thing]` rather than `unless options[:thing].nil?`

spec/cli_spec.rb
```diff
@@ -528,6 +528,81 @@
     end
   end
 
+  context "#set_replication?" do
```
carbonin: I was thinking something more targeted for these.

For example:

```ruby
  it "returns false when cluster node number is not set" do
    subject.options = {:replication => "primary", :password => "password"}
    expect(subject.set_replication?).to be_falsey
  end

  it "returns false when the replication option is set to an invalid value" do
    #more test stuff here
  end
```

This way you don't have to deal with the exception and you are not also testing `#parse` and `#run`.

spec/cli_spec.rb
```diff
@@ -528,6 +528,81 @@
...
+    end
+  end
+
+  context "#replication_params?" do
+    it "should complain if primary-host is not passed while configuring standby replication" do
+      expect { subject.parse(%w(--replication standby --cluster-node-number 2 --password pass")).run }.to raise_error(TrollopEducateSpecError)
+    end
+  end
+
+  context "#set_replication" do
```
carbonin: These look great @rsevilla87 :+1:

PR #18, opened 2017-11-15
Hint for reset database error

ailisp:
Issue:
When reset database with connected appliances, it shows `There are xx existing connections to the database preventing the setup of a database region`, should give a hint suggest user close those evm servers.

BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1512840

\cc @yrudman @gtanzillo 
@miq-bot add-label wip, bug

### DIFF
--- a/lib/manageiq/appliance_console/database_configuration.rb
+++ b/lib/manageiq/appliance_console/database_configuration.rb
@@ -80,7 +80,8 @@ module ApplianceConsole
     end
 
     def create_region
-      ManageIQ::ApplianceConsole::Utilities.bail_if_db_connections("preventing the setup of a database region")
+      hint = "Please stop the EVM server process on all appliances in the region"
+      ManageIQ::ApplianceConsole::Utilities.bail_if_db_connections("preventing the setup of a database region.\n#{hint}")
       log_and_feedback(__method__) do
         ManageIQ::ApplianceConsole::Utilities.rake("evm:db:region", ["--", {:region => region}])
       end


ailisp:
@miq-bot remove-label wip

ailisp:
For the style hint, I think it's OK since I only move the origin one inside a `if ` clause

lib/manageiq/appliance_console/database_configuration.rb
```diff
@@ -46,7 +46,9 @@ def run_interactive
       post_activation
       say("\nConfiguration activated successfully.\n")
     rescue RuntimeError => e
-      puts "Configuration failed#{": " + e.message unless e.class == MiqSignalError}"
+      if e.class != MiqNothingChanged
+        puts "Configuration failed#{": " + e.message unless e.class == MiqSignalError}"
+      end
```
carbonin: I don't think we should change this. Just because we didn't change anything doesn't mean that we didn't fail to create the database.

What we're saying here is that we were not able to do what you requested. I think that should imply that we didn't change the state of the system.

lib/manageiq/appliance_console/database_configuration.rb
```diff
@@ -80,7 +82,8 @@ def create_or_join_region
     end
 
     def create_region
-      ManageIQ::ApplianceConsole::Utilities.bail_if_db_connections("preventing the setup of a database region")
+      hint = "\nPlease stop evm server processes or change their region from all appliances in this region."
```
carbonin: > change their region from all appliances in this region

I don't know what this means. You can't really change an appliances region.

ailisp: Sorry. So the only way to solve this is to stop evm server processes connected to this database?

lib/manageiq/appliance_console/database_configuration.rb
ailisp: It's controversial. @yrudman I think both way is fine if we mentioned on the manual, what do you think?

yrudman: yes, changing existing configuration failed, but i think this message message
 `An internal database already exists.
Choose "Reset Internal Database" to reset the existing installation` 
imply that attempt to change was rejected

this message:
`An internal database already exists.
Choose "Reset Internal Database" to reset the existing installation
Configuration failed
`
sounds like we have broke something, and that was complain from user

carbonin: > An internal database already exists. Choose "Reset Internal Database" to reset the existing installation Configuration failed
> sounds like we have broke something, and that was complain from user

I guess I just disagree. It seems to me that this is more of a misunderstanding rather than a flaw in the wording.

lib/manageiq/appliance_console/database_configuration.rb
carbonin: Yes, all of the evmserverd services on appliances in the region should be stopped.

lib/manageiq/appliance_console/database_configuration.rb
```diff
@@ -46,7 +46,9 @@ def run_interactive
       post_activation
       say("\nConfiguration activated successfully.\n")
     rescue RuntimeError => e
-      puts "Configuration failed#{": " + e.message unless e.class == MiqSignalError}"
+      if e.class != MiqNothingChanged
```
carbonin: Also, don't do this.

If we can't rescue just the exception type we care about then we need to find a better way to not print an error message. Depending on exceptions to implement what is essentially a `goto` is bad enough without doing explicit class type comparisons.

lib/manageiq/appliance_console/database_configuration.rb
yrudman: yes, it is definitely misunderstanding.

lib/manageiq/appliance_console/database_configuration.rb
ailisp: Thanks for this useful hint. Part of your comment I'm not understood. What does `without doing explicit class type comparisons` means?

carbonin: You're comparing the class of the error raise to another class within a rescue block.

This really shouldn't be necessary and this kind of thing usually means that the rest of the code could use improvement/refactoring to remove the need for it.

lib/manageiq/appliance_console/database_configuration.rb
```diff
@@ -80,7 +82,8 @@ def create_or_join_region
     end
 
     def create_region
-      ManageIQ::ApplianceConsole::Utilities.bail_if_db_connections("preventing the setup of a database region")
+      hint = "\nPlease choose \"Stop EVM Server Processes\" in appliance console."
```
carbonin: I think this should mention that this needs to be done for all appliances in the region.

Maybe "Please stop the EVM server process on all appliances in the region"?

lib/manageiq/appliance_console/database_configuration.rb
ailisp: I see. I think i should return a third value for `activate` to indicate nothing changes and check what it returns instead of `raise MiqSignalError unless activate`. But as you discussed with Yuri, you're agree it's misunderstanding. So I'll revert this (and still show `Configuration failed`), right?

carbonin: Yes, I think that the only change for this bug should be the hint about how to remove the database connections.

carbonin:
@ailisp can you add the BZ to the commit message?

ailisp:
@carbonin Sorry, it's added now.

PR #19, opened 2017-11-17
fix appliance console cli dump exception stacks if -k when key exist

ailisp:
Issue:
If run `appliance_console_cli -k` when v2 key exists, it will display error message and dump ruby exception stacks. It should only display error message.
BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1506045

\cc @yrudman @gtanzillo 
@miq-bot add-label bug

### DIFF
--- a/lib/manageiq/appliance_console/cli.rb
+++ b/lib/manageiq/appliance_console/cli.rb
@@ -254,7 +254,8 @@ module ApplianceConsole
     def create_key
       say "#{key_configuration.action} encryption key"
       unless key_configuration.activate
-        raise "Could not create encryption key (v2_key)"
+        say("Could not create encryption key (v2_key)")
+        exit(1)
       end
     end
 


lib/manageiq/appliance_console/cli.rb
```diff
@@ -256,6 +256,8 @@ def create_key
       unless key_configuration.activate
         raise "Could not create encryption key (v2_key)"
       end
+    rescue RuntimeError => e
```
carbonin: This looks like it would make the cli script return a success status to the command line after an error. This is not what we want.

I think a better solution would be to `exit` from the cli after writing the message above. Then we would get the message, exit with an error status, and not catch exceptions which are not this particular error situation (activate returning false).

carbonin:
>  Line 258, Col 9 - Rails/Exit - Do not use exit in Rails applications.

@bdunne @Fryguy can we disable this in repos where we are not necessarily in a rails application?
I feel like it's totally fine to use `exit` when were in something like the console.

bdunne:
That sounds reasonable to me.

PR #20, opened 2017-11-23
cli with --standalone option should work without --internal and --region

ailisp:
Issue: 
In appliance_console_cli, when given options `--standalone`, it should configure database without `--internal` and without `--region`. For example this should be correct:
```bash
appliance_console_cli --username root --password smartvm --dbname vmdb_production --standalone --key --dbdisk /dev/vdb
```
BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1513584

\cc @yrudman @carbonin  @gtanzillo 
@miq-bot add-label wip, bug

### DIFF
--- a/lib/manageiq/appliance_console/cli.rb
+++ b/lib/manageiq/appliance_console/cli.rb
@@ -36,11 +36,11 @@ module ApplianceConsole
     end
 
     def database?
-      hostname
+      options[:standalone] || hostname
     end
 
     def local_database?
-      database? && local?(hostname)
+      database? && (local?(hostname) || options[:standalone])
     end
 
     def certs?
@@ -134,8 +134,12 @@ module ApplianceConsole
         opt :extauth_opts,         "External Authentication Options",   :type => :string
         opt :server,               "Server status",                     :type => :string
       end
-      Trollop.die :region, "needed when setting up a local database" if options[:region].nil? && local_database?
+      Trollop.die :region, "needed when setting up a local database" if region_number_required? && options[:region].nil?
       self
+    end
+
+    def region_number_required?
+      !options[:standalone] && local_database?
     end
 
     def run
--- a/spec/cli_spec.rb
+++ b/spec/cli_spec.rb
@@ -81,6 +81,22 @@ describe ManageIQ::ApplianceConsole::Cli do
     subject.run
   end
 
+  it "should not have run_as_evm_server flag when pass standalone" do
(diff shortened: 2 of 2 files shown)


ailisp:
@miq-bot remove-label wip

lib/manageiq/appliance_console/cli.rb
```diff
@@ -20,7 +20,7 @@ def host
 
     # database hostname
     def hostname
-      options[:internal] ? "localhost" : options[:hostname]
+      options[:standalone] || options[:internal] ? "localhost" : options[:hostname]
```
carbonin: I would rather see this change made in the definition of `#database?`

Instead of making `#hostname` be truthy when `--standalone` is specified, I think it makes more sense to make `#database` truthy when we are given a hostname or when `--standalone` is specified.

ailisp: Sounds good!

ailisp:
@carbonin refactoring and still work

lib/manageiq/appliance_console/cli.rb
```diff
@@ -262,7 +266,7 @@ def create_key
     def install_certs
       say "creating ssl certificates"
       config = CertificateAuthority.new(
-        :hostname => host,
+        :hostname => options[:standalone] ? "localhost" : host,
```
carbonin: Why was this changed?

I think `host` is the appliance hostname, not the database host.

ailisp:
@carbonin sorry about that.

ailisp:
@carbonin Finished specs now. Thanks

spec/cli_spec.rb
```diff
@@ -431,6 +447,48 @@
       end
     end
 
+    context "#local_database?" do
+      it "should return false when no database given" do
+        expect(subject).to receive(:database?).and_return(false)
```
carbonin: I would prefer all of these stubs be done by just setting the options like what you're doing on [line 464](https://github.com/ManageIQ/manageiq-appliance_console/pull/20/files#diff-9f2c191ddaf44048040f9babca4ec4caR464)

ailisp: Thanks, but I'm not agree with this, since we unit test `local_database?`, I think it's better to assume `database?` and other stubs works correctly. For whether `database?` work correctly or not is should be tested in their own context. Is this understanding make sense?

carbonin: Makes sense to me.

PR #21, opened 2017-12-01
fix cli set timezone doesn't work

ailisp:
Issue:
Set timezone in cli doesn't work, because the new timezone parameter isn't sent to `TimezoneConfiguration` class. It's fixed now.
BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1517897

\cc @carbonin, @yrudman, @gtanzillo 

@miq-bot add-label bug

### DIFF
--- a/lib/manageiq/appliance_console/cli.rb
+++ b/lib/manageiq/appliance_console/cli.rb
@@ -272,6 +272,7 @@ module ApplianceConsole
 
     def set_time_zone
       timezone_config = ManageIQ::ApplianceConsole::TimezoneConfiguration.new(options[:timezone])
+      timezone_config.new_timezone = options[:timezone]
       if timezone_config.activate
         say("Timezone configured")
       else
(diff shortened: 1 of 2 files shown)


ailisp:
@miq-bot add-label wip

ailisp:
@miq-bot remove-label wip

lib/manageiq/appliance_console/date_time_configuration.rb
```diff
@@ -112,6 +112,6 @@ def configure_date_time
       logger.error("Failed to apply time configuration: #{e.message}")
       false
     end
-  end # class TimezoneConfiguration
+  end # class DateTimeConfiguration
```
carbonin: Why change this in this PR?

lib/manageiq/appliance_console/timezone_configuration.rb
```diff
@@ -8,8 +8,9 @@ class TimezoneConfiguration
     attr_reader   :current_timzone
     attr_accessor :new_timezone
 
-    def initialize(region_timezone_string)
+    def initialize(region_timezone_string, region_new_timezone_string = "")
```
carbonin: Instead of adding a parameter here, couldn't the cli just set the `new_timzone` using the attr_accessor?

Then in a follow up maybe we can remove the current timezone from the parameter here because it doesn't seem to be used.

PR #22, opened 2017-12-11
Configure database ssl on creation

carbonin:
This PR utilizes changes made in https://github.com/ManageIQ/manageiq-appliance/pull/162 to generate ssl certs for encrypting database traffic.

It also removes conditional logic around setting the `ssl` option in postgresql.conf


carbonin:
Related to https://bugzilla.redhat.com/show_bug.cgi?id=1482697 and https://bugzilla.redhat.com/show_bug.cgi?id=1519079

PR #23, opened 2017-12-13
Add GPT as partition table when disk size are bigger than 2 TB

ailisp:
This is a follow up on https://github.com/ManageIQ/manageiq-appliance_console/pull/11 to fix the issue that can't install CloudForms on a disk larger than 2TB. Because dos partition table support upto 2TB disk, and need to use GPT partition table instead. The fix is almost all done by J. Parrill and specs by Nick LaMuro, and with some comments from Nick Carboni and Jason Frey. I just put them together into this PR.

BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1506997

\cc @yrudman @carbonin @gtanzillo 
@miq-bot add-label bug

### DIFF
--- a/lib/manageiq/appliance_console/logical_volume_management.rb
+++ b/lib/manageiq/appliance_console/logical_volume_management.rb
@@ -43,10 +43,10 @@ module ApplianceConsole
     def create_partition_to_fill_disk
       # Check if you need to create a GPT part table or a MSDOS one in base of
       # max size of partition table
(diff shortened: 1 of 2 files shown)


ailisp:
@miq-bot assign @carbonin

ailisp:
We should close https://github.com/ManageIQ/manageiq-appliance_console/pull/11 after this merged, since that is including in this PR.

PR #24, opened 2017-12-13
Implement set date time in cli

ailisp:
As mentioned in https://github.com/ManageIQ/manageiq-appliance_console/issues/7
Now user can use `--datetime` option to configure date time, (optionally used with other options)
Usage:
To set a given time and date:
`appliance_console_cli --datetime YYYY-MM-DD_HH:MM:SS`
Or to set datetime autosync:
`appliance_console_cli --datetime auto`

\cc @carbonin 
@miq-bot add-label enhancement

### DIFF
--- a/lib/manageiq/appliance_console/cli.rb
+++ b/lib/manageiq/appliance_console/cli.rb
@@ -67,6 +67,10 @@ module ApplianceConsole
       options[:timezone]
     end
 
+    def date_time?
+      options[:datetime]
+    end
+
     def extauth_opts?
       options[:extauth_opts]
     end
@@ -141,6 +145,7 @@ module ApplianceConsole
         opt :iparealm,      "IPA Server realm (optional)", :type => :string
         opt :ca,                   "CA name used for certmonger",       :type => :string,  :default => "ipa"
         opt :timezone,             "Time zone",                         :type => :string
+        opt :datetime,             "Date and time, in YYYY-MM-DDTHH:MM:SS (ISO8601) format", :type => :string
         opt :postgres_client_cert, "install certs for postgres client", :type => :boolean
         opt :postgres_server_cert, "install certs for postgres server", :type => :boolean
         opt :http_cert,            "install certs for http server",     :type => :boolean
@@ -158,8 +163,8 @@ module ApplianceConsole
     def run
       Trollop.educate unless set_host? || key? || database? || tmp_disk? || log_disk? ||
                              uninstall_ipa? || install_ipa? || certs? || extauth_opts? ||
-                             time_zone? || set_server_state? || db_hourly_maintenance? ||
-                             set_replication?
+                             time_zone? || date_time? || set_server_state? ||
+                             db_hourly_maintenance? || set_replication?
       if set_host?
         system_hosts = LinuxAdmin::Hosts.new
         system_hosts.hostname = options[:host]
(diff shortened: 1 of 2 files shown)


ailisp:
@miq-bot add-label enhancement

ailisp:
@miq-bot assign @carbonin

lib/manageiq/appliance_console/cli.rb
```diff
@@ -141,6 +145,7 @@ def parse(args)
         opt :iparealm,      "IPA Server realm (optional)", :type => :string
         opt :ca,                   "CA name used for certmonger",       :type => :string,  :default => "ipa"
         opt :timezone,             "Time zone",                         :type => :string
+        opt :datetime,             "Date and time, in YYYY-MM-DD_HH:MM:SS format", :type => :string
```
carbonin: Couldn't we use a more widespread format here? Maybe [ISO8601](https://en.wikipedia.org/wiki/ISO_8601)?

lib/manageiq/appliance_console/cli.rb
```diff
@@ -280,6 +286,15 @@ def set_time_zone
       end
     end
 
+    def set_date_time
+      date_time_config = ManageIQ::ApplianceConsole::DateTimeConfiguration.new
+      unless options[:datetime] == "auto"
+        date_time_config.manual_time_sync = true
+        date_time_config.new_date, date_time_config.new_time = options[:datetime].split("_")
```
carbonin: Can we do validation against the regular expressions defined in `DateTimeConfiguration`?

When used interactively those are applied during `#ask_questions` so they will not get applied here.

spec/cli_spec.rb
```diff
@@ -680,6 +680,27 @@
     end
   end
 
+  context "#set_date_time" do
+    let(:datetime) { double("date time") }
```
carbonin: I feel like it would be much better do use a real instance here.
This would allow you to make assertions on a real object rather than stubbing setters and getters for attributes.

```ruby
let(:datetime) { ManageIQ::ApplianceConsole::DateTimeConfiguration.new }

before do
  expect(ManageIQ::ApplianceConsole::DateTimeConfiguration).to receive(:new).and_return(datetime)
end
```

ailisp:
@carbonin Looks good now.

lib/manageiq/appliance_console/cli.rb
```diff
@@ -280,6 +286,26 @@ def set_time_zone
...
+      date_time_config = ManageIQ::ApplianceConsole::DateTimeConfiguration.new
+      unless options[:datetime] == "auto"
+        date_time_config.manual_time_sync = true
+        date_time_config.new_date, date_time_config.new_time = options[:datetime].split("T")
+        return unless date_time_valid?(date_time_config)
+      end
+      date_time_config.activate
+    end
+
+    def date_time_valid?(date_time_config)
```
bdunne: Shouldn't this be `ManageIQ::ApplianceConsole::DateTimeConfiguration#valid?`?

ailisp: I think maybe not. `ManageIQ::ApplianceConsole::DateTimeConfiguration` use `ask_for_date/time` and provide the regexp to `highline`, let `highline` validate if date time is valid. So the `DateTimeConfiguration` doens't validate date and time itself. And these validation only needed in `cli`.

PR #25, opened 2017-12-13
Update incorrect comment at module/class end

ailisp:
A trivial fix to update incorrect comment at all module/class end. I go through all files in this repo and this PR includes all of outdated comment of this kind. For class/module definitions that don't end with this kind of comment, keep intact.

\cc @carbonin


ailisp:
@miq-bot assign @carbonin

PR #30, opened 2018-01-16
Check the current database connections before running the restore

carbonin:
This allows us to abort and inform the user before taking any potentially destructive actions

https://bugzilla.redhat.com/show_bug.cgi?id=1534999

### DIFF
--- a/bin/appliance_console
+++ b/bin/appliance_console
@@ -394,6 +394,14 @@ Static Network Configuration
         task_params = []
         uri = nil
 
+        connections = ManageIQ::ApplianceConsole::Utilities.db_connections - 1
+        if connections > 0
+          say("\nDatabase restore failed")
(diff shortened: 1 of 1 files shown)


gtanzillo:
@bdunne You ok with merging this?

bin/appliance_console
```diff
@@ -394,6 +394,14 @@ Static Network Configuration
         task_params = []
         uri = nil
 
+        connections = ManageIQ::ApplianceConsole::Utilities.db_connections - 1
+        if connections > 0
+          say("\nDatabase restore failed")
+          say("\nThere #{connections > 1 ? "are #{connections} connections" : "is 1 connection"} preventing a database restore") if connections > 0
```
Fryguy: why the if condition?... you already know the number of connections is > 0 because of line 398

PR #31, opened 2018-02-02
Remove restore connection checking

carbonin:
As of https://github.com/ManageIQ/manageiq/pull/16942 these checks are implemented in the core application so we don't need to confuse things by doing them here.

This also helps in fixing https://bugzilla.redhat.com/show_bug.cgi?id=1540686 by removing the check which fails when pglogical is active when trying to do a restore.


carbonin:
Marking as WIP until https://github.com/ManageIQ/manageiq/pull/16942 is merged

PR #32, opened 2018-02-02
Remove database maintenance from appliance console

jntullo:
✂️ ✂️ 

No longer needed after https://github.com/ManageIQ/manageiq/pull/16929 and https://github.com/ManageIQ/manageiq/pull/16940

@miq-bot assign @carbonin


abellotti:
more ✂️ fun for @jrafanie

carbonin:
@jntullo Can you also remove these from the [locales files](https://github.com/ManageIQ/manageiq-appliance_console/tree/master/locales)?

kbrock:
nice. would love to get this stuff out of the rake files too (another day)

jrafanie:
> No changes need to be made, but I want to make sure @jrafanie doesn't come along and merge this in too early 😉

Should we make this WIP until then?

carbonin:
> nice. would love to get this stuff out of the rake files too (another day)

@kbrock I don't think any of this is in rake files... At least none of what we'll be using after this change.
Can you point us to what you're referencing? Could be another area that needs to be cleaned up after this change.

kbrock:
@carbonin sorry - it was tangental.

I was thinking of the overlapping code in [[evm_dba.rake]](https://github.com/ManageIQ/manageiq/blob/master/lib/tasks/evm_dba.rake)

PR #33, opened 2018-02-22
Optimize Shared Buffers for PostgreSQL Performance

rvalente:
Pulling in the changes for `shared_buffers` from
https://github.com/ManageIQ/manageiq-appliance/pull/181

cc @carbonin 

Related BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1547161

### DIFF
--- a/lib/manageiq/appliance_console/internal_database_configuration.rb
+++ b/lib/manageiq/appliance_console/internal_database_configuration.rb
@@ -8,8 +8,8 @@ module ApplianceConsole
   class InternalDatabaseConfiguration < DatabaseConfiguration
     attr_accessor :disk, :run_as_evm_server
 
-    DEDICATED_DB_SHARED_BUFFERS = "'3GB'".freeze
-    SHARED_DB_SHARED_BUFFERS = "'2GB'".freeze
+    DEDICATED_DB_SHARED_BUFFERS = "'1536MB'".freeze
+    SHARED_DB_SHARED_BUFFERS = "'1024MB'".freeze
 
     def self.postgres_dir
       PostgresAdmin.data_directory.relative_path_from(Pathname.new("/"))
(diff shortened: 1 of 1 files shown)


lib/manageiq/appliance_console/internal_database_configuration.rb
```diff
@@ -8,8 +8,8 @@ module ApplianceConsole
   class InternalDatabaseConfiguration < DatabaseConfiguration
     attr_accessor :disk, :run_as_evm_server
 
-    DEDICATED_DB_SHARED_BUFFERS = "'1GB'".freeze
-    SHARED_DB_SHARED_BUFFERS = "'128MB'".freeze
+    DEDICATED_DB_SHARED_BUFFERS = "'3GB'".freeze
+    SHARED_DB_SHARED_BUFFERS = "'2GB'".freeze
```
carbonin: I think these values are a bit high.

ManageIQ appliances ship with 6GB of RAM and by default run both the database and the application.

So I'm thinking 500MB for shared and 2GB for dedicated?

rvalente: You're correct, that is high if MIQ ships with 6GB of RAM. Would it make sense to have CloudForms and ManageIQ align with the memory footprint/appliance specs by bumping it to a 4 x 12GB appliance?

JPrause:
@miq-bot add_label blocker

JPrause:
@rvalente can you add the gaprindashvili/yes and fine/yes labels.

carbonin:
@JPrause I think this is more accurately related to https://bugzilla.redhat.com/show_bug.cgi?id=1561112 which is not a blocker and is not marked for backport. Sorry for the confusion.

rvalente:
Agree with @carbonin - this is not a blocker. It would be a nice to have.

JPrause:
@miq-bot remove_label blocker

PR #34, opened 2018-03-13
Validate the encryption key when configuring the database

carbonin:
Requires https://github.com/ManageIQ/manageiq/pull/17149

https://bugzilla.redhat.com/show_bug.cgi?id=1544854


kbrock:
@carbonin Think you can kick this and it will turn 🍏

PR #40, opened 2018-05-03
[WIP] Make cancel less confusing

carbonin:
This should only be needed in the main console file. It wasn't being used in `prompts.rb` at all and it wasn't clear how it was defined in `timezone_configuration.rb`


carbonin:
Adding WIP as I haven't had time to look at this again since the addition of `DatabaseAdmin` and accompanying changes to `Prompts`

PR #42, opened 2018-05-08
Refactors dbrestore action into a class

NickLaMuro:
Refactors the code in `appliance_console` to trigger a database restore, `DatabaseAdmin`, eventually to be used to also handle the database backup actions as well since they are pretty much identical in the questions they would need to ask.  Future PR will implement that feature and will be linked to this one.

Note about `DatabaseAdmin`
------------------------------

This class was created as a example class for my thoughts on how we should structure the `bin/appliance_console` UI moving forward.  Right now, a huge chunk of logic in the `bin/appliance_console` is largely untested, and adding the backup functionality would have been largely a copy and paste effort for what was already there for the `dbrestore` functionality.

Doing this did require moving a few things around (see the first two commits), but no too much should have been changed (if anything) as far as user flow to accommodate.

A separate issue will be opened up in the future to discuss how we can architect this moving forward and will be linked here when that has been created.

Links
-----

* https://bugzilla.redhat.com/show_bug.cgi?id=1535345


locales/container/en.yml
```diff
@@ -16,6 +17,7 @@ en:
     - summary
     - quit
     timezone: Set Timezone
+    dbbackup: Create Database Backup
```
carbonin: This shouldn't be in the container file. We have other solutions for backup and restore there.

carbonin:
Generally I'm all for moving more logic out of `bin/appliance_console`, but I think we should stick with the `#ask_questions` and `#activate` paradigm that every other class in this repo uses.

I think the ultimate goal would be to keep the high level options (I18n) in the main bin file with the "business logic" related to gathering data and executing an operation in a separate class which should also be reusable from the cli in a non-interactive way. My idea was to keep things similar to the workflow [here for changing date and time](https://github.com/ManageIQ/manageiq-appliance_console/blob/3f49d2de27e487e9d6b79d8261d45204be5e5281/bin/appliance_console#L323-L332) with the eventual goal of DRYing up even that logic to where we can just add new classes and add an `i18n => class` hash to add an option to the menu.

With this in mind, I'm confused about what we're gaining by separating out a `UI` namespace and moving the I18n options into a separate file.

Additionally, I think refactoring would be better accomplished in a separate PR rather than being mixed into adding backup functionality.

NickLaMuro:
@carbonin Responding inline:

> ...but I think we should stick with the `#ask_questions` and `#activate` paradigm that every other class in this repo uses.

I have no problem with this, and I basically should be able to change `#run` to `#ask_questions` and take out the `#confirm_and_execute` bit from that and then change it to `#activate`.

That said, the reason I didn't go with that paradigm originally is that I was using `DatabaseConfiguration` as a reference, and was initially confused if it was injected through some means that I was unware of (through something like `HighLine` or the `Prompts` class), but turns out it probably is just a bug:

```ruby
$ bundle exec irb
irb(main):001:0> require 'manageiq-appliance_console'
#=> true
irb(main):002:0> db_config = ManageIQ::ApplianceConsole::DatabaseConfiguration.new
#=> #<ManageIQ::ApplianceConsole::DatabaseConfiguration:0x007fdb29cf32b0 @adapter="postgresql">
irb(main):003:0> db_config.methods.grep(/ask_questsions/)
#=> []
irb(main):004:0> ca = ManageIQ::ApplianceConsole::CertificateAuthority.new
#=> #<ManageIQ::ApplianceConsole::CertificateAuthority:0x007fdb28ba97e8 @ca_name="ipa">
irb(main):005:0> ca.methods.grep(/ask_questions/)
#=> [:ask_questions]
irb(main):006:0>
```

And in the `DatabaseConfiguration` class "called" [here](https://github.com/ManageIQ/manageiq-appliance_console/blob/3f49d2d/lib/manageiq/appliance_console/database_configuration.rb#L39).

> With this in mind, I'm confused about what we're gaining by separating out a UI namespace and moving the I18n options into a separate file.

Okay, this is kind of coupled with the things you talked about in the paragraph prior to it, but I am going to answer this in two parts:  the `I18n` stuff first, and then the `UI` namespacing business.

For the `I18n` portion, that was strictly out of necessity.  I had to make a call to `I18n.t` [here](https://github.com/NickLaMuro/manageiq-appliance_console/blob/b98fa6a9/lib/manageiq/appliance_console/ui/database_admin.rb#L130) so I could basically print the header.  Now, without getting to far ahead of myself, basically the idea behind putting this here inside the class is it seem stupid to pass the string to the class by adding another arg to the initialize method.

So to the other question of "why the `UI` namespace?", it stems from me my "firmly held belief" that bin files should basically be this:

```ruby
#!/usr/bin/env ruby

require 'manageiq-appliance_console/ui'

ManageIQ::ApplianceConsole::UI.run
```

But you would have hated me if I did that in one go... so I didn't.

That said, I had a plan/idea of having a `UI` class that was a "base class", that holds the basic scaffolding the others will inherit, like the `#ask_questions` and `#activate` methods, and then simple things like [the `say("#{selection}\n\n")`](https://github.com/ManageIQ/manageiq-appliance_console/blob/3f49d2d/bin/appliance_console#L323) stuff that is basically repeated for every choice in `bin/appliance_console`.

Things like `I18n` would be part of a `ManageIQ::ApplianceConsole::UI#setup` method that would be triggered by the internals of `.run`, among many of the other parts of the first part of `bin/appliance_console` prior to the main ui loop (signal trapping, gathering appliance info, setting constants, etc).  Basically, when all of this is in `bin/appliance_console`, though, it isn't testable.

What I would agree with you and your concerns with this `UI` name spacing is that it kind of mixes both the `#activate` with the ui specific stuff which is the `#ask_questions` bit.  Ideally, I probably should have made both a `DatabaseAdmin` that is initialized as part of `UI::DatabaseAdmin`, that is configured in `#ask_questions` and called out to in `#activate`.  This could then be reused by the `cli` as well, but it seemed like `dbrestore` wasn't used in the CLI so I skipped this step for now.  The idea, though, was to eventually compartmentalize the things that are specific to `bin/appliance_console` under the `UI` namespace umbrella.

> Additionally, I think refactoring would be better accomplished in a separate PR rather than being mixed into adding backup functionality.

That's fine, and I simply truncate the last two commits from this PR and add them into a separate PR that builds off this one.

That said, I figured without that last two commits the building off the existing class wouldn't have made as much sense as to "why" I was doing this.  Also, I didn't want to start with basically copying and pasting the `dbrestore` portion with no tests to implement the "dbbackup" feature, only rip it out in another PR to integrate this refactoring after (a flawed plan because of duplicate work in hindsight, but how I originally thought this would go if I started with the "feature first").

locales/container/en.yml
NickLaMuro: This is one of those situations where I wasn't sure, but saw when messing with the `I18n` stuff that there were two versions of this.  So I figured I would add it in both and you could yell at me if I got it wrong (as you did), instead of me forgetting about not adding it and then needing to add it in a follow up/bug fix.

tl; dr  Thanks!  Will remove.

carbonin:
> I didn't go with that paradigm originally is that I was using DatabaseConfiguration as a reference

Unfortunately `DatabaseConfiguration` isn't a great example because if the weird inheritance thing it does. I would take a look at one of the other classes for a reference, [key configuration](https://github.com/ManageIQ/manageiq-appliance_console/blob/3f49d2de27e487e9d6b79d8261d45204be5e5281/lib/manageiq/appliance_console/key_configuration.rb) is another good example.

> For the I18n portion, that was strictly out of necessity. I had to make a call to I18n.t here so I could basically print the header.

Okay, this is fine for now then.

> `ManageIQ::ApplianceConsole::UI.run`

That's awesome and I definitely agree that's where we want to go. Allso UI is a great name for the class that runs the main console logic, but I don't think your new class belongs in a `UI` namespace as most if it's important contents should be shared between the UI and CLI. I think that when we get to the point of creating a `UI` class it can just use the same classes that `bin/appliance_console` is using now. I don't see the need for a namespace at all.

> it seemed like dbrestore wasn't used in the CLI so I skipped this step for now

There's an issue from way back tracking things that need to be added to the CLI (https://github.com/ManageIQ/manageiq-appliance_console/issues/7). It would be great to have backup in the CLI from the start, but no need to add restore here.

-------------------------------------

So at a high level it sounds like you want to pull the existing restore code out of `bin/appliance_console` into a new class, `DatabaseAdmin`, add the new backup feature to that class, and do it in a way that makes sense fore when we make a bigger refactoring change to move all the main menu logic into a `UI` class.

I'm good with that as long as backup makes it into the CLI, `DatabaseAdmin` isn't namespaced in `UI`, and that the public API for the class is `ask_questions` and `activate`. What do you think?

NickLaMuro:
> I'm good with that as long as backup makes it into the CLI, DatabaseAdmin isn't namespaced in UI, and that the public API for the class is ask_questions and activate. What do you think?

Works for me.  I might "fight" you more on the `UI` namespace eventually, but that can be saved for another time, and for the purposes here it definitely doesn't need to happen right now.  I think I still need to elaborate and state my case/rational better ("what... more :books: !?"), but I think that can be done in github issue that would flesh out my vision a little bit and provide some more concrete context and examples for further discussion.

For now, I think it makes perfect sense to go ahead with your plan since it make more sense with the current architecture and won't confuse new comers  👍 .  I will trim down this PR and open up another to add the `backup` functionality just for continuity, and it won't take much more work.

Thanks!

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -0,0 +1,119 @@
+require 'manageiq/appliance_console/errors'
+
+module ManageIQ
+  module ApplianceConsole
+    class DatabaseAdmin < HighLine
```
carbonin: Why is this needed? I don't think we do this anywhere else.

NickLaMuro: _@NickLaMuro realizes he is going to be writing another_ :book: _to explain this... \*sigh\*_

tl; dr  #Reasons™

To start off:

Actually [we do](https://github.com/ManageIQ/manageiq-appliance_console/blob/3f49d2de/spec/prompts_spec.rb#L22), it is just not done within the codebase normally when we run `bin/appliance_console`.

In `bin/appliance_console` in "production", by contrast, [import](https://github.com/ManageIQ/manageiq-appliance_console/blob/3f49d2de/bin/appliance_console#L14-L15), [include](https://github.com/ManageIQ/manageiq-appliance_console/blob/3f49d2de/bin/appliance_console#L21), and even [configure](https://github.com/ManageIQ/manageiq-appliance_console/blob/3f49d2de/bin/appliance_console#L35-L36) `HighLine` within the `bin/appliance_console` to make the DSL methods "globally available" via `$terminal` [[ref]](https://github.com/JEG2/highline/blob/42f1e7c3/lib/highline/import.rb#L11).

For various reasons, #thisSucks™ outside of a quick single file script because it means we can't test the configuration we set there, and I don't have access to that in the tests unless I include `highline/import` within the tests, which would probably mess with the `prompts_spec.rb` that I wasn't willing to deal with.

We have also basically been punting on the `HighLine` DSL in the specs as well, so anything that actually used the DSL in it's class (`certificate_authority.rb` for example) was either stubbing it or just not running those methods.

* * *

Now doing this has it's drawbacks.  Technically speaking, we are using a different instance of `HighLine` here that isn't `$terminal`, so the `$terminal.wrap_at=` and `$terminal.page_at=` calls are not respected in this case.  My plan was to get out of using the global state for this with the `UI.run` stuff, so for now we are "split brained" with the addition of this which might not be the best.  There might also be some weirdness with in practice when there is both the `$terminal` DSL and the inherited one exist at the same time, but Ruby should just favor the local methods over the ones that are forwarded on from `Kernel`.

That said, I don't think there is much that takes advantage `page_at` or `wrap_at` of this in `DatabaseAdmin`, so I think we are fine, for now (any of the major printing we might see probably will be probably from `AwesomeSpawn`, so that probably won't be managed by `HighLine`).  But moving forward, this is part of the reason I wanted to split components the parts that need the `HighLine` DSL ("UI" specific) from the parts that are doing the `#activate` code and would be shared with the CLI (will add this to my case in the "GH issue I have yet to create").

carbonin: But given that every I/O method we use here with the exception of `#say` is actually defined on `Prompts` would it be better to have that class inherit from `Highline` then get the methods from the include?

I feel like it's easier to understand when there is only one place that the I/O methods are coming from.

NickLaMuro: @carbonin well, `Prompts` is a `module`... so not sure that would work... 😞

NickLaMuro: > I feel like it's easier to understand when there is only one place that the I/O methods are coming from.

But yes, 100% agree with you on this, just not sure how much of this I should try and tackle here.  Open to suggestions.

carbonin: Ah right, damn.

bin/appliance_console
```diff
@@ -408,56 +397,8 @@ Static Network Configuration
...
-          rake_success = ManageIQ::ApplianceConsole::Utilities.rake(task, task_params)
-          if rake_success && delete_agreed
-            say("\nRemoving the database restore file #{DB_RESTORE_FILE}...")
-            File.delete(DB_RESTORE_FILE)
-          elsif !rake_success
-            say("\nDatabase restore failed. Check the logs for more information")
-          end
-        end
-        press_any_key
+        db_admin = ManageIQ::ApplianceConsole::UI::DatabaseAdmin.new
```
carbonin: I think we need to remove the `::UI` namespace here.

NickLaMuro: Oh, nevermind.  Sorry, I thought I didn't remove the namespace at all, but instead I left this in when I shouldn't have.  Good catch!

Also... this is why I hate `bin` files being so large...

NickLaMuro:
@carbonin we good to go here?

PR #44, opened 2018-05-25
Add dbbackup action

NickLaMuro:
This is built off of https://github.com/ManageIQ/manageiq-appliance_console/pull/42 and adds new functionality to trigger `evm:db:backup:local` and `evm:db:backup:remote`, with identical options for file location as `restore` (Local file, NFS, and SMB).

This basically slightly modifies the `DatabaseAdmin` class to support an `action` accessor, which allows the `bin/appliance_console` to choose between doing a `:restore` or a `:backup`.  From there, it is simply choosing the right file options and such from the menu in the same fashion as for `restore`.

Links
-----

* https://github.com/ManageIQ/manageiq-appliance_console/pull/42


NickLaMuro:
@carbonin Sorry for the 3 or so rebases, but I think I got this where `codeclimate` isn't going to yell at me, and I am happy with myself.

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -88,29 +89,56 @@ def ask_smb_file_options
...
+
       def setting_header
-        say("#{I18n.t("advanced_settings.dbrestore")}\n\n")
+        say("#{I18n.t("advanced_settings.db#{action}")}\n\n")
+      end
+
+      private
+
+      def processing_message
+         say("\n#{action == :restore ? "Restoring the database" : "Running Database backup to #{uri}"}...")
```
NickLaMuro: This looks like trash, but I think when I end up adding the `:dump` action into the mix, I will end up making this look a little nicer then.

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -88,29 +89,56 @@ def ask_smb_file_options
...
-          end
+        if allowed_to_execute?
+          processing_message
+          run_rake
         end
         press_any_key
       end
 
+      def allowed_to_execute?
+        if action == :restore
```
carbonin: Can we make this a guard?

```ruby
return true unless action == :restore
say("\nNote: A database restore cannot be undone.  The restore will use the file: #{uri}.\n")
agree("Are you sure you would like to restore the database? (Y/N): ")
```

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -88,29 +89,56 @@ def ask_smb_file_options
...
+          say("\nNote: A database restore cannot be undone.  The restore will use the file: #{uri}.\n")
+          agree("Are you sure you would like to restore the database? (Y/N): ")
+        else
+          true
+        end
+      end
+
+      def file_menu_args
+        [
+          action == :restore ? "Restore Database File" : "Backup File Location",
```
carbonin: Reading through this class I've seen the file to be used for a restore called the "backup file" a few times so this was a bit confusing.
Would something like "Backup Result File Name" make more sense here? (not sold on the name, looking for suggestions ... @jrafanie don't you love naming things?! :wink: )

spec/database_admin_spec.rb
```diff
@@ -18,294 +18,705 @@
...
+          expect(subject.uri).to         eq(example_uri)
+          expect(subject.task).to        eq("evm:db:backup:remote")
+          expect(subject.task_params).to eq(expected_task_params)
+        end
       end
     end
 
-    context "when @backup_type not is LOCAL_FILE" do
-      let(:uri) { described_class::DB_RESTORE_FILE }
+    describe "#ask_to_delete_backup_after_restore" do
```
carbonin: Is this in the "for DB backup" context? I think this is restore related.

locales/container/en.yml
```diff
@@ -5,6 +5,7 @@ en:
   advanced_settings:
     menu_order:
     - timezone
+    - dbbackup
```
carbonin: We have a separate backup solution for containers. We should probably re-consider the options in this list as a whole, but this one definitely doesn't belong.

lib/manageiq/appliance_console/database_admin.rb
carbonin: Yeah, this is fine for now :+1:

carbonin:
Also cc @roliveri @jerryk55 I think this will make it a bit easier to add the other backup/restore targets.

jerryk55:
Thanks @carbonin!  I'll take a look.

lib/manageiq/appliance_console/database_admin.rb
NickLaMuro: Oh, I like this better.  Thanks!

spec/database_admin_spec.rb
NickLaMuro: Right, but because of how I have things, this get's fired every time.  That is why I am confirming a `no-op` with the spec below:

```ruby
it "no-ops" do
  # ...
end
```

lib/manageiq/appliance_console/database_admin.rb
NickLaMuro: I mean, I am not sold on anything myself, so if you would prefer that, I can go with that option as well.

locales/container/en.yml
NickLaMuro: Butts... this was carry over from your last review, and I was cherry-picking in what it was in the previous diff:

https://github.com/ManageIQ/manageiq-appliance_console/compare/3f49d2de27e487e9d6b79d8261d45204be5e5281...NickLaMuro:rfe-db-dump-from-appliance-console-menu-orig

So I didn't take the time to look over what was reviewed there... sorry.  Will make the change.

lib/manageiq/appliance_console/database_admin.rb
NickLaMuro: @carbonin Would `"Backup Output File Name"` be better?

carbonin: Yeah, that sounds good :+1:

spec/database_admin_spec.rb
carbonin: Yup it would help if I read the specs :sweat_smile: Carry on

NickLaMuro: No worries, and I appreciate the attention to detail.  It also admittedly is a weird spec, so no shame in the mixup here.

PR #45, opened 2018-05-26
Add dbdump action

NickLaMuro:
This continues to built off of #42 and #44, and adds new functionality to trigger `evm:db:dump:local` and `evm:db:dump:remote`, again with identical options for file location as restore (Local file, NFS, and SMB).

A slight change up from #44 is that this also adds in a warning for the "Dump" action, since it is not recommended/support to use `pg_dump` for database backups.  To ensure that users are aware of the difference, we are printing a warning when ever this action is executed.  Besides that, it behaves almost identically to "Backup" at the moment.

Links
-----

* Partially addresses https://bugzilla.redhat.com/show_bug.cgi?id=1535345
* Relies on https://github.com/ManageIQ/manageiq/pull/17483 being merged
* Diff specific to the changes between this PR and #44:  [diff](https://github.com/NickLaMuro/manageiq-appliance_console/compare/rfe-db-dump-from-appliance-console-menu-part-2...rfe-db-dump-from-appliance-console-menu-part-3)

### DIFF
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -18,6 +18,14 @@ module ManageIQ
         Example: 'mydomain.com/user'
       PROMPT
 
+      DB_DUMP_WARNING = <<-WARN.strip_heredoc
+        WARNING:  This is not the recommended and supported way of running a
+        database backup, and is strictly meant for exporting a database for
+        support/debugging purposes!
+
+
+      WARN
+
       attr_reader :action, :backup_type, :task, :task_params, :delete_agree, :uri
 
       def initialize(action = :restore, input = $stdin, output = $stdout)
@@ -29,6 +37,7 @@ module ManageIQ
 
       def ask_questions
         setting_header
+        say(DB_DUMP_WARNING) if action == :dump
         ask_file_location
       end
 
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -103,7 +103,7 @@ module ManageIQ
 
       def file_menu_args
         [
-          action == :restore ? "Restore Database File" : "Backup Output File Name",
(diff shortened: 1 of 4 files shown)


lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -9,33 +9,37 @@ class DatabaseAdmin < HighLine
...
         Example: 'mydomain.com/user'
       PROMPT
 
-      attr_accessor :backup_type, :task, :task_params, :delete_agree, :uri
+      DB_DUMP_WARNING = <<-WARN.strip_heredoc
+        WARNING:  This is not the recommended and supported way of running a
+        database backup, and is strictly meant for exporting a database for
+        support/debugging purposes!
+
+        Please use the 'Backup' functionality instead.
```
carbonin: I think this last line should be qualified in some way or just removed. Out of context it feels like you should never be using this option.

NickLaMuro: Yeah, this was meant to be a draft, and I figured there would have been some thoughts around the actual copy of it.

I did want to point them to the #Correct™ way of doing things, but didn't want to fall into my trap of getting to wordy and write another 📖 .  I am fine with deleting it as well, though I guess I would prefer to keep it for the aforementioned reasoning, but don't really have a good qualifier in mind.  My only though was:

```
Please use the 'Backup' functionality instead if you intend
to create a backup to restore in application failure and
disaster recovery scenarios.
```

carbonin:
The PR this is based on was merged. Can you rebase this so it's easier to see the changes specific to this PR?

NickLaMuro:
@carbonin oh yeah, good point!

Give me one second.

NickLaMuro:
Opps... fat fingered that...

Anyway, rebase done.

carbonin:
Good to go outside of https://github.com/ManageIQ/manageiq/pull/17483 not being merged.

PR #47, opened 2018-06-05
Allow specifying table exclusions in Database Dump action

NickLaMuro:
This pull request exposes the functionality of the `--exclude-table-data` flag in `pg_dump`, which was added to `manageiq` and `manageiq-gems-pending` in the following pull requests respectively:

* https://github.com/ManageIQ/manageiq/pull/17483
* https://github.com/ManageIQ/manageiq-gems-pending/pull/351

In the user experience would look something like this:

```
Would you like to exclude tables in the dump? (Y/N): y
To exclude tables from the dump, enter them in a space separated
list.  For example:

    > metrics_* vim_performance_states event_streams

Enter the tables to exclude: |metrics_* vim_performance_states event_streams| event_streams
```

Links
-----

* Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1586186
* https://github.com/ManageIQ/manageiq/pull/17483
* https://github.com/ManageIQ/manageiq-gems-pending/pull/351

### DIFF
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -39,6 +39,7 @@ module ManageIQ
         setting_header
         say(DB_DUMP_WARNING) if action == :dump
         ask_file_location
+        ask_for_tables_to_exclude_in_dump
       end
 
       def activate
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -96,6 +96,26 @@ module ManageIQ
         end
       end
 
+      def ask_for_tables_to_exclude_in_dump
+        if action == :dump && should_exclude_tables?
+          say(<<-PROMPT.strip_heredoc)
+
+            To exclude tables from the dump, enter them in a space separated
+            list.  For example:
+
+                > metrics_* vim_performance_states event_streams
+
+          PROMPT
+          table_excludes = ask_for_many("table",
+                                        "tables to exclude",
+                                        "metrics_* vim_performance_states event_streams",
+                                        255,
+                                        Float::INFINITY)
+
+          @task_params.last[:"exclude-table-data"] = table_excludes
+        end
+      end
+
       def confirm_and_execute
         if allowed_to_execute?
           processing_message
@@ -124,6 +144,12 @@ module ManageIQ
       end
 
       private
+
+      def should_exclude_tables?
+        ask_yn?("Would you like to exclude tables in the dump") do |q|
+          q.readline = true
+        end
+      end
 
       def local_file_prompt
         if action == :restore
--- a/lib/manageiq/appliance_console/prompts.rb
+++ b/lib/manageiq/appliance_console/prompts.rb
@@ -63,6 +63,7 @@ module ApplianceConsole
         q.default = default if default
         q.validate = ->(p) { (p.blank? && default) || %w(y n).include?(p.downcase[0]) }
         q.responses[:not_valid] = "Please provide yes or no."
+        yield q if block_given?
       end.downcase[0] == 'y'
     end
 
--- a/spec/database_admin_spec.rb
+++ b/spec/database_admin_spec.rb
@@ -304,31 +304,56 @@ describe ManageIQ::ApplianceConsole::DatabaseAdmin, :with_ui do
           expect_output ""
         end
       end
+    end
 
-      describe "#confirm_and_execute" do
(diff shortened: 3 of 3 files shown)


NickLaMuro:
Code Climate failure is failing on something that I modified, and really isn't worth fixing at this point in my opinion.

carbonin:
This looks good, I'm just not sure about the defaults for the exclude. I would lean towards not excluding any tables by default. Would want some more opinions before we merge this.

@jdeubel @jrafanie @dmetzger57 what do you think?

dmetzger57:
I'm good with the tables chosen as the defaults to be excluded (they are the ones typically bloating a dump and are not needed), and at a higher level in having a default set of tables on this option. However, I defer to support on whether a default would be good or bad in the wild.

jdeubel:
I would say by default have "vim_performance_states and event_streams" as these tables can take up a bit of space and are very rarely needed from my experience. For the metrics* tables I feel the customer is just going to hit enter since it is quick and easy. We should leave it up to  (support) to tell them what other tables are needed (such as metrics) as the customer might not know.

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -96,6 +97,26 @@ def ask_to_delete_backup_after_restore
         end
       end
 
+      def ask_for_tables_to_exclude_in_dump
+        if action == :dump && should_exlcude_tables?
```
jrafanie: typo: should_exlcude_tables -> should_exclude_tables

+2 points for making the typo consistently in the method name.  😉 

Don't worry, I type `destory` instead of `destroy`.

jrafanie:
I know we don't want feature creep but showing them which tables are the top x in size, might make it easier for them to know which to exclude.

I don't have a strong opinion on default tables either way. We can always ask for the missing tables in the event we're missing tables we need.  What @jdeubel said is good with me.

lib/manageiq/appliance_console/database_admin.rb
NickLaMuro: `performance` is the bane of my existence usually... ends up as `preformance`...

Good catch.  Thanks!

NickLaMuro: > +2 points for making the typo consistently in the method name. 😉

Autocomplete...

NickLaMuro:
> I know we don't want feature creep but showing them which tables are the top x in size, might make it easier for them to know which to exclude.

I am going to say no to this for now, possible future feature though.  Reason being is we really rely on the getting DB connection info from the application, and the `appliance_console` is just a thin wrapper around that in most cases.  I feel like adding a rake tasks, or something similar that takes seconds to run (because of application bootup time specifically) to get this feature in would be a deterrent more than a help (since most people, as mentioned earlier, would just want to hit `ENTER`).

> This looks good, I'm just not sure about the defaults for the exclude...

Just for the record, the tables I chose were not picked at random, but plucked from the previous ticket and discussion:

https://bugzilla.redhat.com/show_bug.cgi?id=1535345#c12

Specifically to @jdeubel 's comment, the metrics tables from `metrics_*` are just the ones that are collected (and purged) on a daily (every 4 hours?) basis (so `metrics_00`, `metrics_01`, etc.), which I assumed in most cases was not the issue that has gotten support involved at this point.  These are not the long term storage metric tables.

The other reason I wanted to have the `metrics_*` in there it is a small hint/example that globbing is possible with this, which I think is important to avoid having them avoid as much typing as possible.

* * *

That last bit about "avoiding typing" was a big goal for this, and I hope it influences the overall decision here.  I put a `y/n` question right before this so if excluding was not wanted at all, then the default excludes wouldn't be injected.  But if exclusions were desired, I wanted the "95% case" to be something that could just be another `ENTER` keystroke away, without making this feature too complex.

Anyway, that is my `</2cents>`.  Would have put it in the original PR description, but it probably wouldn't have been read anyway then.  😭

jrafanie:
> Specifically to @jdeubel 's comment, the metrics tables from metrics_* are just the ones that are collected (and purged) on a daily (every 4 hours?) basis (so metrics_00, metrics_01, etc.), which I assumed in most cases was not the issue that has gotten support involved at this point. These are not the long term storage metric tables.
> 
> The other reason I wanted to have the metrics_* in there it is a small hint/example that globbing is possible with this, which I think is important to avoid having them avoid as much typing as possible.

Those seem like good reasons to include `metrics_*` in the default exclusion list.

jdeubel:
@jrafanie I'm ok with that. It all depends on the case for what we need but I would assume they would have our guidance on what to exclude anyways.

PR #48, opened 2018-06-27
Add support to Enable OpenID-Connect in the appliance console.

jvlcek:
https://www.pivotaltracker.com/n/projects/1610127/stories/121849185
https://www.pivotaltracker.com/n/projects/1610127/stories/121849185

The change introduced by this PR will allow the appliance console support of the OpenID Connect protocol.

This is a WIP PR for now as we only support OpenID in the pods now with the configmap generator. Standalone appliance apache config does not support OpenID (i.e. we don't have the separate Apache templates & the needed manual steps in http://manageiq.org/docs/reference/ under Authentication).

The change introduced by this PR is reliant on the following work for full OpenID Connect support:

[#16495](https://github.com/ManageIQ/manageiq/pull/16495)
ManageIQ/manageiq-ui-classic#2855

### DIFF
--- a/lib/manageiq/appliance_console/external_auth_options.rb
+++ b/lib/manageiq/appliance_console/external_auth_options.rb
@@ -92,11 +92,11 @@ module ApplianceConsole
       end
 
       if params.include?("/authentication/saml_enabled=true")
-        params = remove_oidc(params)
+        remove_oidc(params)
       elsif params.include?("/authentication/oidc_enabled=true")
-        params = remove_saml(params)
+        remove_saml(params)
       elsif params.include?("/authentication/oidc_enabled=false") || params.include?("/authentication/saml_enabled=false")
-        params = provider_type_none(params)
+        provider_type_none(params)
       end
     end
 
--- a/lib/manageiq/appliance_console/external_auth_options.rb
+++ b/lib/manageiq/appliance_console/external_auth_options.rb
@@ -81,7 +81,7 @@ module ApplianceConsole
       if update_hash.present?
         say("\nUpdating external authentication options on appliance ...")
         params = update_hash.collect { |key, value| "#{key}=#{value}" }
-        params = enable_provider_type!(params)
+        params = configure_provider_type!(params)
         result = ManageIQ::ApplianceConsole::Utilities.rake_run("evm:settings:set", params)
         raise parse_errors(result).join(', ') if result.failure?
       end
@@ -93,27 +93,27 @@ module ApplianceConsole
       false
     end
 
-    def enable_provider_type!(params)
+    def configure_provider_type!(params)
       if params.include?("/authentication/saml_enabled=true")
-        provider_type_saml(params)
+        configure_saml!(params)
       elsif params.include?("/authentication/oidc_enabled=true")
-        provider_type_oidc(params)
+        configure_oidc!(params)
       elsif params.include?("/authentication/oidc_enabled=false") || params.include?("/authentication/saml_enabled=false")
-        provider_type_none(params)
+        configure_none!(params)
       end
     end
 
-    def provider_type_saml(params)
+    def configure_saml!(params)
       params << "/authentication/oidc_enabled=false"
       params << "/authentication/provider_type=saml"
     end
 
-    def provider_type_oidc(params)
+    def configure_oidc!(params)
       params << "/authentication/saml_enabled=false"
       params << "/authentication/provider_type=oidc"
     end
 
-    def provider_type_none(params)
+    def configure_none!(params)
       params << "/authentication/oidc_enabled=false"
       params << "/authentication/saml_enabled=false"
       params << "/authentication/provider_type=none"
--- a/lib/manageiq/appliance_console/external_auth_options.rb
+++ b/lib/manageiq/appliance_console/external_auth_options.rb
@@ -81,7 +81,7 @@ module ApplianceConsole
       if update_hash.present?
         say("\nUpdating external authentication options on appliance ...")
         params = update_hash.collect { |key, value| "#{key}=#{value}" }
-        params = enable_provider_type(params)
+        params = enable_provider_type!(params)
         result = ManageIQ::ApplianceConsole::Utilities.rake_run("evm:settings:set", params)
         raise parse_errors(result).join(', ') if result.failure?
       end
@@ -93,7 +93,7 @@ module ApplianceConsole
       false
     end
 
-    def enable_provider_type(params)
+    def enable_provider_type!(params)
       if params.include?("/authentication/saml_enabled=true")
         provider_type_saml(params)
       elsif params.include?("/authentication/oidc_enabled=true")
--- a/lib/manageiq/appliance_console/external_auth_options.rb
+++ b/lib/manageiq/appliance_console/external_auth_options.rb
@@ -87,27 +87,40 @@ module ApplianceConsole
     end
 
     def never_enable_saml_and_oidc(params)
-      if (params.include? "/authentication/oidc_enabled=true") && (params.include? "/authentication/saml_enabled=true")
+      if params.include?("/authentication/oidc_enabled=true") && params.include?("/authentication/saml_enabled=true")
         say("\nWARNING: Both SAML and OIDC can not be enable. SAML will be enabled ...")
       end
 
-      if params.include? "/authentication/saml_enabled=true"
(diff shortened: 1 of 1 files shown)


jvlcek:
@miq-bot add_label authentication

jvlcek:
@miq-bot add_label gaprindashvili/no

jvlcek:
@miq-bot add_label wip

lib/manageiq/appliance_console/external_auth_options.rb
```diff
@@ -79,11 +80,49 @@ def update_configuration(update_hash = nil)
...
         params = update_hash.collect { |key, value| "#{key}=#{value}" }
+        params = never_enable_saml_and_oidc(params)
         result = ManageIQ::ApplianceConsole::Utilities.rake_run("evm:settings:set", params)
         raise parse_errors(result).join(', ') if result.failure?
       end
     end
 
+    def never_enable_saml_and_oidc(params)
+      if params.include?("/authentication/oidc_enabled=true") && params.include?("/authentication/saml_enabled=true")
+        say("\nWARNING: Both SAML and OIDC can not be enable. SAML will be enabled ...")
```
carbonin: Typo `s/enable/enabled/`

lib/manageiq/appliance_console/external_auth_options.rb
```diff
@@ -79,11 +80,49 @@ def update_configuration(update_hash = nil)
       if update_hash.present?
         say("\nUpdating external authentication options on appliance ...")
         params = update_hash.collect { |key, value| "#{key}=#{value}" }
+        params = never_enable_saml_and_oidc(params)
         result = ManageIQ::ApplianceConsole::Utilities.rake_run("evm:settings:set", params)
         raise parse_errors(result).join(', ') if result.failure?
       end
     end
 
+    def never_enable_saml_and_oidc(params)
```
carbonin: Methods that alter the parameter passed in should end in a `!`

lib/manageiq/appliance_console/external_auth_options.rb
```diff
@@ -79,11 +80,49 @@ def update_configuration(update_hash = nil)
...
+      if params.include?("/authentication/saml_enabled=true")
+        params = remove_oidc(params)
+      elsif params.include?("/authentication/oidc_enabled=true")
+        params = remove_saml(params)
+      elsif params.include?("/authentication/oidc_enabled=false") || params.include?("/authentication/saml_enabled=false")
+        params = provider_type_none(params)
+      end
+    end
+
+    def remove_oidc(params)
```
carbonin: These should also all end in `!` to show that the parameter is being altered.

lib/manageiq/appliance_console/external_auth_options.rb
```diff
@@ -79,11 +80,49 @@ def update_configuration(update_hash = nil)
...
       end
     end
 
+    def never_enable_saml_and_oidc(params)
+      if params.include?("/authentication/oidc_enabled=true") && params.include?("/authentication/saml_enabled=true")
+        say("\nWARNING: Both SAML and OIDC can not be enable. SAML will be enabled ...")
+      end
+
+      if params.include?("/authentication/saml_enabled=true")
+        params = remove_oidc(params)
```
carbonin: If you are editing the hash passed in, you don't need to re-assign.

lib/manageiq/appliance_console/external_auth_options.rb
```diff
@@ -79,11 +80,49 @@ def update_configuration(update_hash = nil)
...
         say("\nUpdating external authentication options on appliance ...")
         params = update_hash.collect { |key, value| "#{key}=#{value}" }
+        params = never_enable_saml_and_oidc(params)
         result = ManageIQ::ApplianceConsole::Utilities.rake_run("evm:settings:set", params)
         raise parse_errors(result).join(', ') if result.failure?
       end
     end
 
+    def never_enable_saml_and_oidc(params)
+      if params.include?("/authentication/oidc_enabled=true") && params.include?("/authentication/saml_enabled=true")
```
carbonin: I'm not sure what the user experience is here, but why can this even happen? Shouldn't we ask for the user to specify only one external auth type? Or check for this as they are selecting the options rather than assuming one way or the other after the fact?

Either way I think it would be better to fail here rather than just picking one of the options they said they want to configure.

jvlcek: The logic in `external_auth_options.rb` expects all the options to be a toggle of a `true` or `false` value. The new authentication setting `provider_type` is used by the UI  to toggle `oidc_enabled` and `saml_enabled`.

After speaking with @abellotti about this, the agreed approach is to get the needed changes in that will make it work even if not elegant and avoid a larger refactoring, since the fate of OIDC on non-podified appliances is not finalized.

Because the appliance_console loops allowing the user to reselect options the params array could have many combinations. Without reworking `external_auth_options.rb` the solution I presented ensures correct results.

That said, I could error out if both `oidc_enabled` and `saml_enabled` are true.

jvlcek: Oh and thanks for taking a look! @carbonin

carbonin: Okay, then I think a `validate!` which will error if there is an incompatible combination selected is a better experience than possibly configuring something the user may not have intended.

jvlcek: @carbonin 

The only incompatible combination it so enable both oidc and saml. I will add a validate! for that and error out.

Thanks!

lib/manageiq/appliance_console/external_auth_options.rb
```diff
@@ -79,11 +81,44 @@ def update_configuration(update_hash = nil)
...
         params = update_hash.collect { |key, value| "#{key}=#{value}" }
+        params = enable_provider_type(params)
         result = ManageIQ::ApplianceConsole::Utilities.rake_run("evm:settings:set", params)
         raise parse_errors(result).join(', ') if result.failure?
       end
     end
 
+    def validate_provider_type
+      return true unless @updates["/authentication/oidc_enabled"] == true && @updates["/authentication/saml_enabled"] == true
+      say("\Error: Both SAML and OIDC can not be enable ...")
```
carbonin: Typo .. should be "enabled"

lib/manageiq/appliance_console/external_auth_options.rb
```diff
@@ -79,11 +81,44 @@ def update_configuration(update_hash = nil)
...
+      false
+    end
+
+    def enable_provider_type(params)
```
carbonin: Since you're altering the passed-in hash here, I think this method should be named `enable_provider_type!`

lib/manageiq/appliance_console/external_auth_options.rb
```diff
@@ -79,11 +81,44 @@ def update_configuration(update_hash = nil)
...
       end
     end
 
+    def validate_provider_type
+      return true unless @updates["/authentication/oidc_enabled"] == true && @updates["/authentication/saml_enabled"] == true
+      say("\Error: Both SAML and OIDC can not be enabled ...")
+      false
+    end
+
+    def enable_provider_type(params)
```
carbonin: This should be named with a `!` because you're altering the parameter passed into the method. (same with all the `provider_type_*` methods

lib/manageiq/appliance_console/external_auth_options.rb
```diff
@@ -79,11 +81,44 @@ def update_configuration(update_hash = nil)
...
+      if params.include?("/authentication/saml_enabled=true")
+        provider_type_saml(params)
+      elsif params.include?("/authentication/oidc_enabled=true")
+        provider_type_oidc(params)
+      elsif params.include?("/authentication/oidc_enabled=false") || params.include?("/authentication/saml_enabled=false")
+        provider_type_none(params)
+      end
+    end
+
+    def provider_type_saml(params)
```
carbonin: I think I would prefer these have a method name that uses a verb as it's altering the hash. A noun method feels more like it's returning a value retrieved from somewhere else.

Also all of these should be named with a `!` for the same reason as above.

jvlcek:
@carbonin and @abellotti Can the failing codeclimate _Complexity_ issue be waved?
_**Method ask_questions has a Cognitive Complexity of 6 (exceeds 5 allowed). Consider refactoring**_

This PR and https://github.com/ManageIQ/manageiq/pull/16567 need to be merged together.

PR #49, opened 2018-06-29
Upgrade to repmgr version 4

carbonin:
This PR makes all the necessary changes to be compatible with the newest major version of [repmgr](https://repmgr.org/docs/4.0/index.html)

In particular this PR does the following:

- Removes the cluster name and associated logic (this is no longer used by repmgr)
- Renames the config file keys
- Alters the primary registration command from `master` to `primary`
- Removes the logic around adding the schema to the search path
- Slightly alters the `promote` and `follow` commands
- Adds configuration to use `systemctl` to alter the state of the postgres cluster

The upgrade in combination with some enhancements here will solve the following bus and tracking issues:
https://bugzilla.redhat.com/show_bug.cgi?id=1418080
https://www.pivotaltracker.com/story/show/135779733
https://www.pivotaltracker.com/story/show/141523501

### DIFF
--- a/lib/manageiq/appliance_console/database_replication_primary.rb
+++ b/lib/manageiq/appliance_console/database_replication_primary.rb
@@ -30,27 +30,8 @@ module ApplianceConsole
     def activate
       say("Configuring Primary Replication Server...")
       create_config_file(primary_host) &&
-        initialize_primary_server &&
+        run_repmgr_command(REGISTER_CMD) &&
         write_pgpass_file
-    end
-
-    def initialize_primary_server
-      run_repmgr_command(REGISTER_CMD) &&
-        add_repmgr_schema_to_search_path
-    end
-
-    def add_repmgr_schema_to_search_path
-      schema_name = "repmgr"
-      begin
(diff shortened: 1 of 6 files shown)


carbonin:
This is WIP because we need to coordinate merge, gem releases, and new RPMs, but I think it's ready for review.

carbonin:
@jvlcek can you take a look? I think you worked on some of this originally, right?

lib/manageiq/appliance_console/database_replication_primary.rb
```diff
@@ -30,28 +30,9 @@ def confirm
...
+        run_repmgr_command(REGISTER_CMD) &&
         write_pgpass_file
     end
-
-    def initialize_primary_server
-      run_repmgr_command(REGISTER_CMD) &&
-        add_repmgr_schema_to_search_path
-    end
-
-    def add_repmgr_schema_to_search_path
```
gtanzillo: Do we need to remove or change the old search path on upgrade?

carbonin:
Un-WIPing this. We can merge this now, but can't release a new console version until we have the rest of the repmgr v4 stuff in place.

PR #50, opened 2018-07-21
Adds file splitting support to dump/backup commands

NickLaMuro:
~~**NOTE:**  Currently built off of https://github.com/ManageIQ/manageiq-appliance_console/pull/51~~ **Update:**  This has been rebased and this branch updated accordingly.

This exposes the `:byte_count` flag that is added to the `evm:db:backup` and `evm:db:dump` tasks here:

* https://github.com/ManageIQ/manageiq/pull/17894

The user experience will look something like the following:

```
Would you like to split the dump output into multiple parts? (Y/N): y
Enter the byte size to split by: |500M| 750M
```

Links
-----

* Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1586176
* Relies on the following to be merged to function:
  - [x] https://github.com/ManageIQ/manageiq-gems-pending/pull/361
  - [x] https://github.com/ManageIQ/manageiq/pull/17894
  - [x] [ManageIQ/manageiq-appliance_console#51](https://github.com/ManageIQ/manageiq-appliance_console/pull/51)

### DIFF
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -42,6 +42,7 @@ module ManageIQ
         say(DB_DUMP_WARNING) if action == :dump
         ask_file_location
         ask_for_tables_to_exclude_in_dump
+        ask_to_split_up_output
       end
 
       def activate
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -147,6 +147,12 @@ module ManageIQ
         end || true
       end
 
+      def ask_to_split_up_output
+        if action == :dump && should_split_output?
+          @task_params.last[:byte_count] = ask_for_string("byte size to split by", "500M")
+        end || true
+      end
+
       def confirm_and_execute
         if allowed_to_execute?
           processing_message
@@ -178,6 +184,12 @@ module ManageIQ
 
       def should_exclude_tables?
         ask_yn?("Would you like to exclude tables in the dump") do |q|
+          q.readline = true
+        end
+      end
+
+      def should_split_output?
+        ask_yn?("Would you like to split the #{action} output into multiple parts") do |q|
           q.readline = true
         end
       end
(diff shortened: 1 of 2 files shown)


NickLaMuro:
Might change how we are handling file splitting, so setting this as `[WIP]` for now.  Most likely, not much is going to change with this, but the underlying pre-requisite PRs needed for this are most likely going to change, so this can't be merged until those are sorted out.

See https://github.com/ManageIQ/manageiq/pull/17798#issuecomment-410339880 for some further details.

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -147,6 +148,12 @@ def ask_for_tables_to_exclude_in_dump
         end || true
       end
 
+      def ask_to_split_up_output
+        if action != :restore && should_split_output?
```
carbonin: I think we should only support this for the `:dump` action on the first pass since restore from split backups won't work and it wasn't really part of the original RFE.

carbonin: I guess what I'm saying is that we shouldn't give the user the option to backup using splits if we can't restore with splits. I'm fine with the option to dump using splits because the intention was never to restore it.

NickLaMuro: > I guess what I'm saying is that we shouldn't give the user the option to backup using splits if we can't restore with splits.

I guess that is fair.

I figured that someone would eventually request the splitting functionality for backups as well anyway, specifically for `s3`, since there limits to file size.

* * *

Please note that I am also partially arguing this point simply because I don't want to delete the appliance console integration tests that I have written that already support running against and validating split based backups...

(deep down... I am really a simple human being really...)

carbonin: Definitely not arguing against creating split backups at some point, but I think we need split restore at that point as well. If we leave this as is, some user will be rather unhappy when we have to tell them that they can't restore a backup that we let them create.

NickLaMuro: Welp, I can do that.  Despite not wanting to comment out/remove some of my personal tests, I think I agree with that line of thinking.

Will try and take care of that code change shortly.

PR #51, opened 2018-07-23
Fixes to DatabaseAdmin commands

NickLaMuro:
Fixes the following in `ManageIQ::ApplianceConsole::DatabaseAdmin`

* `LOCAL_FILE_VALIDATOR` for creating a `:dump` and a `:backup`, since it would check for the files existance when creating a new database dump/backup.
* The default for a local dump file is now `/tmp/evm_db.dump`, not `/tmp/evm_db.backup`
* Fix `ask_questions` so it returns true (breaks how things are written in `bin/appliance_console` otherwise)
* Fixes questions for `SMB` and `NFS` file locations to allow specifying the name of the `:remote_file_name` properly.

Testing
-------

I have been using the following gist to test these changes:

https://gist.github.com/NickLaMuro/8438015363d90a2d2456be650a2b9bc6

The tests are actually built around testing https://github.com/ManageIQ/manageiq-appliance_console/pull/50, but these changes are needed for any of that to work, and that PR is built on top of this one, so it has been helpful in making sure all the necessary changes are actually doing what I want.

### DIFF
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -61,11 +61,7 @@ module ManageIQ
       end
 
       def ask_local_file_options
-        validator = LOCAL_FILE_VALIDATOR if action == :restore
-        default   = action == :dump ? DB_DEFAULT_DUMP_FILE : DB_RESTORE_FILE
-
-        @uri = just_ask(local_file_prompt, default, validator, "file that exists")
-
+        @uri         = just_ask(*filename_prompt_args)
         @task        = "evm:db:#{action}:local"
         @task_params = ["--", {:local_file => uri}]
       end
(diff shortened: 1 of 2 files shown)


NickLaMuro:
@carbonin I think so.  I have been actively testing it with the gist above to catch all of my mistakes found in this PR.  I still haven't tested against the "table exclusion" portion of things, but I can open up a second PR if that ends up having issues since this PR already has a bunch of fixes associated with it.

carbonin:
Okay, @NickLaMuro if we want to cut a console version do you know if there are any other issues that need to get in for testing? I think we should get a version of this to QE soon.

carbonin:
Even though there was no BZ associated with this I'm going to label it as a bug fix because it is fixing issues caused by previous enhancement PRs in this area.

NickLaMuro:
@carbonin Probably should get #50 merged in first, at least, which has these changes baked in already.

PR #54, opened 2018-08-09
Enable DB Restore from S3

jerryk55:
Add an option to the DB Restore step to choose AWS S3 as the source of the database file from
which to restore.  Prompt for the AWS Region and credentials as well.

Requires https://github.com/ManageIQ/manageiq/pull/17827 as well which modifies
the rake db restore from remote step.

@NickLaMuro @carbonin @roliveri please review and merge when appropriate.

This is the last of the PRs for the RFE for S3 DB Backups in https://bugzilla.redhat.com/show_bug.cgi?id=1513520.

### DIFF
--- a/bin/appliance_console
+++ b/bin/appliance_console
@@ -67,7 +67,8 @@ include ManageIQ::ApplianceConsole::Prompts
 RESTORE_LOCAL   = "Local file".freeze
 RESTORE_NFS     = "Network File System (NFS)".freeze
 RESTORE_SMB     = "Samba (SMB)".freeze
-RESTORE_OPTIONS = [RESTORE_LOCAL, RESTORE_NFS, RESTORE_SMB, ManageIQ::ApplianceConsole::CANCEL].freeze
+RESTORE_S3      = "AWS S3 (S3)".freeze
+RESTORE_OPTIONS = [RESTORE_LOCAL, RESTORE_NFS, RESTORE_SMB, RESTORE_S3, ManageIQ::ApplianceConsole::CANCEL].freeze
 
 # Restart choices
 RE_RESTART  = "Restart".freeze
--- a/bin/appliance_console
+++ b/bin/appliance_console
@@ -67,8 +67,7 @@ include ManageIQ::ApplianceConsole::Prompts
 RESTORE_LOCAL   = "Local file".freeze
 RESTORE_NFS     = "Network File System (NFS)".freeze
 RESTORE_SMB     = "Samba (SMB)".freeze
-RESTORE_S3      = "AWS S3 (S3)".freeze
-RESTORE_OPTIONS = [RESTORE_LOCAL, RESTORE_NFS, RESTORE_SMB, RESTORE_S3, ManageIQ::ApplianceConsole::CANCEL].freeze
+RESTORE_OPTIONS = [RESTORE_LOCAL, RESTORE_NFS, RESTORE_SMB, ManageIQ::ApplianceConsole::CANCEL].freeze
 
 # Restart choices
 RE_RESTART  = "Restart".freeze
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -105,7 +105,7 @@ module ManageIQ
         @uri         = ask_for_uri(*remote_file_prompt_args_for("s3"))
         user         = just_ask(access_key_prompt)
         pass         = ask_for_password("Secret Access Key for #{user}")
-        region       = ask_for_region("Amazon Region for database file")
+        region       = just_ask("Amazon Region for database file", "us-east-1")
 
         @task        = "evm:db:#{action}:remote"
         @task_params = [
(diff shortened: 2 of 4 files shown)


bin/appliance_console
```diff
@@ -67,7 +67,8 @@ include ManageIQ::ApplianceConsole::Prompts
 RESTORE_LOCAL   = "Local file".freeze
 RESTORE_NFS     = "Network File System (NFS)".freeze
 RESTORE_SMB     = "Samba (SMB)".freeze
-RESTORE_OPTIONS = [RESTORE_LOCAL, RESTORE_NFS, RESTORE_SMB, ManageIQ::ApplianceConsole::CANCEL].freeze
+RESTORE_S3      = "AWS S3 (S3)".freeze
+RESTORE_OPTIONS = [RESTORE_LOCAL, RESTORE_NFS, RESTORE_SMB, RESTORE_S3, ManageIQ::ApplianceConsole::CANCEL].freeze
```
carbonin: I don't see this being used in this file ... Are we sure these need to be here anymore?

cc @NickLaMuro

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -8,7 +8,8 @@ class DatabaseAdmin < HighLine
       LOCAL_FILE     = "Local file".freeze
       NFS_FILE       = "Network File System (NFS)".freeze
       SMB_FILE       = "Samba (SMB)".freeze
-      FILE_OPTIONS   = [LOCAL_FILE, NFS_FILE, SMB_FILE, CANCEL].freeze
+      S3_FILE        = "Amazon S3 (S3)".freeze
```
carbonin: Why is this a different string from the `RESTORE_S3` constant above? If we need the one in `bin/appliance_console` then these should definitely be the same.

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -19,6 +20,11 @@ class DatabaseAdmin < HighLine
         Example: 'mydomain.com/user'
       PROMPT
 
+      ACCESS_KEY_PROMPT = <<-PROMPT.strip_heredoc.chomp
```
carbonin: S3 is the only method that will use this, right? If that's the case, can this string live in the `#ask_s3_file_options` method?

bin/appliance_console
jerryk55: Looks like no, but I didn't want to remove them as part of this PR, and felt that I should be consistent.  In hindsight probably just should have figured it out and removed them.

lib/manageiq/appliance_console/database_admin.rb
jerryk55: This is the right one...

lib/manageiq/appliance_console/database_admin.rb
jerryk55: Will do...

jerryk55:
@carbonin your comments applied.  Note that I only removed the new constant added to bin/appliance_console - not the existing entries there.  I leave those as an exercise for a separate PR.

jerryk55:
Something really strange happened here.  My previous commits disappeared and were merged together somehow. I think I inadvertantly switched to my old laptop to make this commit and push - looking into it.

lib/manageiq/appliance_console/prompts.rb
```diff
@@ -137,6 +138,10 @@ def ask_for_password(prompt, default = nil)
       pass == "********" ? (default || "") : pass
     end
 
+    def ask_for_region(prompt, default = "us-east-1")
```
carbonin: I don't think this should be in a separate `Prompts` method. I would prefer you use `just_ask("Amazon Region for database file", "us-east-1")` in the `DatabaseAdmin` class

lib/manageiq/appliance_console/prompts.rb
```diff
@@ -18,6 +18,7 @@ module Prompts
     SAMPLE_URLS = {
       'nfs' => 'nfs://host.mydomain.com/exported/my_exported_folder/db.backup',
       'smb' => 'smb://host.mydomain.com/my_share/daily_backup/db.backup',
+      's3'  => 's3://mybucket/my_subdirectory/daily_backup/db.backup',
```
carbonin: Not strictly for this PR, but why is this constant defined here? The only place I see it used is in `DatabaseAdmin` outside of the method `sample_url` here, but that's never called outside of specs ... 

@NickLaMuro ?

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -94,6 +96,29 @@ def ask_smb_file_options
...
+        region       = ask_for_region("Amazon Region for database file")
+
+        @task        = "evm:db:#{action}:remote"
+        @task_params = [
+          "--",
+          {
+            :uri          => uri,
+            :uri_username => user,
+            :uri_password => pass,
+            :uri_region   => region
```
carbonin: I think this needs to change to `aws_region` right?

jerryk55:
@carbonin agreed and done.

lib/manageiq/appliance_console/prompts.rb
NickLaMuro: It was here before I was working on `DatabaseAdmin`, so I didn't move it.

PR #55, opened 2018-08-13
Fix reference to MiqSignalError in ensure_key_configured

yrudman:
Moving `ensure_key_configured` out of `ManageIQ::ApplianceConsole` scope introduced `ensure_key_configured: uninitialized constant MiqSignalError (NameError)`

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1596127

@miq-bot add-label bug

### DIFF
--- a/bin/appliance_console
+++ b/bin/appliance_console
@@ -42,7 +42,7 @@ def ensure_key_configured
     else
       say("\nEncryption key not configured.")
       press_any_key
-      raise MiqSignalError
+      raise ManageIQ::ApplianceConsole::MiqSignalError
     end
   end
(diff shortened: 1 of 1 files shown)


carbonin:
There look like there could be quite a few of these. @yrudman I'm okay with fixing just this one in this PR, but do you want to try to tackle the others as well?

yrudman:
@carbonin  other refs to `MiqSignalError` seems to be in correct namespace, could you point to some other example

carbonin:
Ah you're right, because they're used within namespaced classes. Got it.

Looks good

PR #57, opened 2018-08-16
Adds :remote_file_name to task_params for DatabaseAdmin#ask_s3_file_options

NickLaMuro:
The way that backup/dump is handled for for the rake tasks, it expects that the `--uri` is the location of the "share" (in this case, the s3 bucket), and the `--remote-file-name` is used to define the file name:

https://github.com/ManageIQ/manageiq/blob/5c204228/lib/evm_database_ops.rb#L118-L120

Made more confusing I am sure with how the questions are asked in the DatabaseAdmin lib for the URI.  Admittedly I wrote a lot of this, but based it off of what was already there for the `restore` task, so I reserve some blame to what was there previously and me trying to stick to that interface...  (still my fault).

Without it, if you try and pass just the `--uri`.  For example `s3://bucketname/path/to/backup.tar.gz`) without a `--remote-file-name`, the result will end up as:

```
s3://buckname/path/to/backup.tar.gz/db_backup/miq_backup_20180101_000000
````

This will fix things by allowing the rake task to properly name the file, instead of using what was the default, and using the URI to set a base folder.

* * *

In addition, I have added some tests around this code, and had to slightly re-arrange some questions to allow those specs to work (see first commit message for more info).

### DIFF
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -102,21 +102,22 @@ module ManageIQ
           Example: 'amazon_aws_user'
         PROMPT
 
+        @filename    = just_ask(*filename_prompt_args) unless action == :restore
         @uri         = ask_for_uri(*remote_file_prompt_args_for("s3"))
         region       = just_ask("Amazon Region for database file", "us-east-1")
         user         = just_ask(access_key_prompt)
         pass         = ask_for_password("Secret Access Key for #{user}")
 
+        params = {
+          :uri          => uri,
+          :uri_username => user,
+          :uri_password => pass,
+          :aws_region   => region
+        }
+        params[:remote_file_name] = filename if filename
+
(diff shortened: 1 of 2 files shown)


NickLaMuro:
@miq-bot assign @carbonin 

cc @jerryk55 

Hopefully this change makes sense.  Ran into when adding s3 tests to my [`appliance_console` integration tests](https://gist.github.com/NickLaMuro/8438015363d90a2d2456be650a2b9bc6), and noticed that what was in the appliance console wouldn't work with the code for:

```conosle
$ rake evm:db:backup:remote
````

Sorry I didn't notice this in https://github.com/ManageIQ/manageiq-appliance_console/pull/54

NickLaMuro:
@miq-bot add_label bug

jerryk55:
Yeah slightly confused by the comment indicating what the filename will be if you just pass --uri.  I'm pretty sure that isn't what happens.

PR #58, opened 2018-08-16
Fix:  Make s3 uri paths optional

NickLaMuro:
Built off of #57 (merge that first)

Based on the format provided in the sample URL for s3:

```
s3://mybucket/path/to/file
```

and how the implementation is coded up in `MiqS3Session`, the bucket name is plucked from the host (which can just be the bucket name, and not the hostname like `bucketname.s3.amazonaws.com`).  In the case of s3, if no path is provided to the URL, it means that the resulting directory is top level, which is should be valid.

This fix simply allows that functionality by updating the `ask_for_uri` validator to all no paths in the uri conditionally (based on an argument option) and hooks up the `#ask_s3_file_options` method to use this new flag.  All other uses of `#ask_for_uri` should still function as they did.

Links
-----

* https://github.com/ManageIQ/manageiq-appliance_console/pull/57
* Diff:  https://github.com/NickLaMuro/manageiq-appliance_console/compare/fix-s3-options-for-database-admin...fix-s3-uri-validation

### DIFF
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -103,7 +103,7 @@ module ManageIQ
         PROMPT
 
         @filename    = just_ask(*filename_prompt_args) unless action == :restore
-        @uri         = ask_for_uri(*remote_file_prompt_args_for("s3"))
+        @uri         = ask_for_uri(*remote_file_prompt_args_for("s3"), :optional_path => true)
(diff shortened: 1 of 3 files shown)


NickLaMuro:
@miq-bot assign @carbonin
@miq-bot add_label bug

cc @jerryk55

carbonin:
@NickLaMuro can you rebase this so that the diff doesn't contain the changes from #57

PR #60, opened 2018-08-16
Update summary to report OIDC or SAML configured

jvlcek:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1438546

The appliance console falsely reported that external authentication was not configured when it actually was for OpenID Connect or SAML.

This PR adds support to check of the httpd remote user configuration file used for each of these external authentication options has been created.

It is possible that the user could configure then and then use the UI to un-configure external authentication and not remove the httpd remote user file(s) and the appliance console would
erroneously report external authentication was still configured.

However this is also the case with the current functionality for the sssd.conf file. So this PR
does no more harm than what already exists and it adds the benefit if reporting SAML and/or
OIDC configured when they are.

### DIFF
--- a/lib/manageiq/appliance_console/external_httpd_authentication/external_httpd_configuration.rb
+++ b/lib/manageiq/appliance_console/external_httpd_authentication/external_httpd_configuration.rb
@@ -221,10 +221,10 @@ module ApplianceConsole
 
     def self.config_status
       fetch_ipa_configuration("ipa_server") ||
-      fetch_sssd_domain ||
-      fetch_oidc_configured? ||
-      fetch_saml_configured? ||
-      "not configured"
+        fetch_sssd_domain                   ||
+        fetch_oidc_configured?              ||
+        fetch_saml_configured?              ||
+        "not configured"
     end
 
     def self.ipa_client_configured?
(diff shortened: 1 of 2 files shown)


jvlcek:
@miq-bot add_label bug

jvlcek:
@miq-bot add_label authentication

jvlcek:
@miq-bot assign @abellotti

lib/manageiq/appliance_console/external_httpd_authentication/external_httpd_configuration.rb
```diff
@@ -243,6 +248,14 @@ def self.fetch_sssd_domain
       config_file_read(SSSD_CONFIG)[/\[domain\/(.*)\]/, 1] if File.exist?(SSSD_CONFIG)
     end
 
+    def self.fetch_saml_configured?
```
carbonin: This method naming is strange. A method with a question mark usually indicates `true` or `false` return value but this method is returning a status string. I think this might be better if it was called `saml_status`

Same applies to the `fetch_oidc_configured?` method

jvlcek: Good point. My initial approach was to have these methods return `true` or `false`. I forgot to rename them when I decided on this approach.

jvlcek: Thank you @carbonin

PR #62, opened 2018-08-21
Don't return empty params when when not altering provider type.

jvlcek:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1619662

[PR 48](https://github.com/ManageIQ/manageiq-appliance_console/pull/48) introduced a failure where if the provider type was not being modified the modification params were inadvertently cleared. So no external authentication params could be altered unless the provider type settings
were also being altered.

The fix is to simply return the unaltered params list from method `configure_provider_type!` when no provider type modifications are being done.

This impacted toggling both SSO and Local Login even though the BZ only mentions SSO.

### DIFF
--- a/spec/external_auth_options_spec.rb
+++ b/spec/external_auth_options_spec.rb
@@ -1,0 +1,67 @@
+describe ManageIQ::ApplianceConsole::ExternalAuthOptions do
+  subject { described_class.new }
+  let(:result) { double(@spec_name, :failure? => false) }
+  let(:rake_set) { "evm:settings:set" }
+
+  before do
+    @spec_name = File.basename(__FILE__).split(".rb").first.freeze
+    allow(subject).to receive(:say)
+  end
+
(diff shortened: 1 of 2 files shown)


spec/external_auth_options_spec.rb
```diff
@@ -0,0 +1,67 @@
+describe ManageIQ::ApplianceConsole::ExternalAuthOptions do
+  subject { described_class.new }
+  let(:result) { double(@spec_name, :failure? => false) }
+  let(:rake_set) { "evm:settings:set" }
+
+  before do
+    @spec_name = File.basename(__FILE__).split(".rb").first.freeze
```
carbonin: You don't really need this. You can just give the double a string name like "RakeResult" and it will work fine.

jvlcek: @carbonin Thank you for looking this over. Will do.

JoeV

PR #64, opened 2018-09-19
Adds FTP support to backup/dump/restore

NickLaMuro:
Built off of https://github.com/ManageIQ/manageiq-appliance_console/pull/50

Adds questions for asking for FTP based options for appliance console DatabaseAdmin tasks (restore, backup, and dump)

Links
-----
* https://bugzilla.redhat.com/show_bug.cgi?id=1586187
* Relies on the following to be merged to function:
  - https://github.com/ManageIQ/manageiq-appliance_console/pull/50

### DIFF
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -9,7 +9,8 @@ module ManageIQ
       NFS_FILE       = "Network File System (NFS)".freeze
       SMB_FILE       = "Samba (SMB)".freeze
       S3_FILE        = "Amazon S3 (S3)".freeze
-      FILE_OPTIONS   = [LOCAL_FILE, NFS_FILE, SMB_FILE, S3_FILE, CANCEL].freeze
+      FTP_FILE       = "File Transfer Protocol (FTP)".freeze
+      FILE_OPTIONS   = [LOCAL_FILE, NFS_FILE, SMB_FILE, S3_FILE, FTP_FILE, CANCEL].freeze
 
       DB_RESTORE_FILE      = "/tmp/evm_db.backup".freeze
       DB_DEFAULT_DUMP_FILE = "/tmp/evm_db.dump".freeze
@@ -59,6 +60,7 @@ module ManageIQ
         when NFS_FILE   then ask_nfs_file_options
         when SMB_FILE   then ask_smb_file_options
         when S3_FILE    then ask_s3_file_options
+        when FTP_FILE   then ask_ftp_file_options
         when CANCEL     then raise MiqSignalError
         end
       end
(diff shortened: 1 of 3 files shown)


carbonin:
Does FTP *need* file splitting? I guess I'm asking if we should even allow FTP targets for backup if we can't split for backup?

NickLaMuro:
@carbonin I think you are seeing the commits from #50 mixed in with this one.

There is nothing different with file splitting for FTP that is different with any other file storage provider.

NickLaMuro:
I will take care of rebasing and getting those extra commits out right now though.

carbonin:
Yes, I understand that the implementation of file splitting doesn't change, but I'm asking more about the requirements of FTP servers as a whole.

Let me try a different phrasing:

Can we stream a backup to an FTP server if the user says "no" to file splitting?

carbonin:
Also, is this ready to un-wip or is there more work to be done with this one?

NickLaMuro:
> Also, is this ready to un-wip or is there more work to be done with this one?

You caught me at the beginning of my day, and #50 was not merged when I called it a day yesterday.

Currently have a weird quirk with getting the anonymous FTP tests to work [here](https://gist.github.com/NickLaMuro/8438015363d90a2d2456be650a2b9bc6), but I tried it manually and it is working fine, so it is something to do with input/output parsing with the console most likely.

> Can we stream a backup to an FTP server if the user says "no" to file splitting?

Short answer:  yes

These seem like more of questions that should have been asked in the implementation of `MiqFileStorage` than here, so I am honestly kind of confused they are coming up now.

But again, EVERYTHING under the `MiqFileStorage` umbrella should be streaming it's file uploads (and downloads ideally, a bug I am still in the process of fixing), which then enables EVERYTHING to do file splitting.

That said, it is provider specific, but I have made it a point to make sure we are always streaming uploads with everything implemented.  For FTP:

https://github.com/ManageIQ/manageiq-gems-pending/blob/5ad64326/lib/gems/pending/util/object_storage/miq_ftp_storage.rb#L41

I was able to use `IO.copy_stream`, which pretty much makes streaming trivial for the most part, and since Socket (what `conn` is under the hood in that method) inherits from IO, this works just fine.  For Swift, however (from the branch I did):

https://github.com/ManageIQ/manageiq-gems-pending/compare/master...NickLaMuro:swift-storage-as-object-storage#diff-4e18367370f0acfd0aca45d85b6aecabR47

https://github.com/ManageIQ/manageiq-gems-pending/compare/master...NickLaMuro:swift-storage-as-object-storage#diff-0d98542e21431fa34f986a4ba977f828R26

I had to do things a bit more manually in `excon`, but streaming still happens, just a bit more in ruby than in "C land".  But this means the case is also the same for Swift as well, assuming that `.upload_single` is implemented in that fashion.

But even if we aren't streaming, we are still calling `upload_single` for a single file, which will stream (assuming the provider implemented that method properly).

PR #65, opened 2018-09-24
Anonymous FTP upload

kbrock:
This will allow customers to upload files to support sites like `dropbox.redhat.com`

To add this option, add to `en.yml`:

```yml
  database_admin:
    menu_order:
    - local
    - ftp://ftp.example.com/incoming/999999-db.backup
    local: Local file
    prompts:
      ftp.example.com:
        filename_text: "The case number dash (-) filename. (e.g.: 12345-db.backup)"
        filename_validator: "^[0-9]{4,}-..*"
```

https://bugzilla.redhat.com/show_bug.cgi?id=1632433

### DIFF
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -5,12 +5,11 @@ module ManageIQ
     class DatabaseAdmin < HighLine
       include ManageIQ::ApplianceConsole::Prompts
 
-      LOCAL_FILE     = "Local file".freeze
-      NFS_FILE       = "Network File System (NFS)".freeze
-      SMB_FILE       = "Samba (SMB)".freeze
-      S3_FILE        = "Amazon S3 (S3)".freeze
-      FTP_FILE       = "File Transfer Protocol (FTP)".freeze
-      FILE_OPTIONS   = [LOCAL_FILE, NFS_FILE, SMB_FILE, S3_FILE, FTP_FILE, CANCEL].freeze
+      LOCAL_FILE = "local".freeze
+      NFS_FILE   = "nfs".freeze
+      SMB_FILE   = "smb".freeze
+      S3_FILE    = "s3".freeze
+      FTP_FILE   = "ftp".freeze
 
       DB_RESTORE_FILE      = "/tmp/evm_db.backup".freeze
       DB_DEFAULT_DUMP_FILE = "/tmp/evm_db.dump".freeze
@@ -28,13 +27,6 @@ module ManageIQ
 
 
       WARN
-
-      SAMPLE_URLS = {
-        'nfs' => 'nfs://host.mydomain.com/exported/my_exported_folder/db.backup',
-        'smb' => 'smb://host.mydomain.com/my_share/daily_backup/db.backup',
-        's3'  => 's3://mybucket/my_subdirectory/daily_backup/db.backup',
-        'ftp' => 'ftp://host.mydomain.com/path/to/daily_backup/db.backup'
-      }
 
       attr_reader :action, :backup_type, :task, :task_params, :delete_agree, :uri, :filename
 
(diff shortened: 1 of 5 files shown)


lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -138,6 +138,27 @@ def ask_ftp_file_options
...
+        @filename    = just_ask(*filename_prompt_args) unless action == :restore
+        sample_case  = uri.split("/").last
+        require 'uri'
+        hostname     = URI(uri).host
+        prompt_text  = I18n.t("database_admin.prompt")[hostname.to_sym] || "Target filename (e.g.: #{sample_case})"
+        prompt_regex = I18n.t("database_admin.validator")[hostname.to_sym]
+        validator    = ->(x) { x.to_s =~ /#{prompt_regex}/ } if prompt_regex
+        uri_filename = just_ask(prompt_text, nil, validator)
+        @uri         = uri.gsub(sample_case, uri_filename)
+        pass         = just_ask("email address (to use as the anonymous ftp password)")
```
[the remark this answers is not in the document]
kbrock: This is the standard anonymous ftp thing. But I think skipping username/password makes things easier

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -138,6 +138,27 @@ def ask_ftp_file_options
         @task_params = ["--", params]
       end
 
+      def ask_anonymous_file_options(uri)
+        @filename    = just_ask(*filename_prompt_args) unless action == :restore
+        sample_case  = uri.split("/").last
+        require 'uri'
+        hostname     = URI(uri).host
+        prompt_text  = I18n.t("database_admin.prompt")[hostname.to_sym] || "Target filename (e.g.: #{sample_case})"
+        prompt_regex = I18n.t("database_admin.validator")[hostname.to_sym]
```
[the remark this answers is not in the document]
kbrock: Spoke with @jcarter12 and he felt this level of education was better handled by documentation.

I did take a stab at the different hierarchy

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -55,13 +54,14 @@ def activate
...
       def ask_file_location
-        case @backup_type = ask_with_menu(*file_menu_args)
-        when LOCAL_FILE then ask_local_file_options
-        when NFS_FILE   then ask_nfs_file_options
-        when SMB_FILE   then ask_smb_file_options
-        when S3_FILE    then ask_s3_file_options
-        when FTP_FILE   then ask_ftp_file_options
-        when CANCEL     then raise MiqSignalError
+        @backup_type = ask_with_menu(*file_menu_args) do |menu|
+          menu.choice(CANCEL) { |_| raise MiqSignalError }
```
carbonin: I wonder if this is something we should just add to the end of every menu? Unlikely that it should be added to this PR, but this made me think of it.

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -185,10 +200,21 @@ def allowed_to_execute?
         agree("Are you sure you would like to restore the database? (Y/N): ")
       end
 
+      def file_options
+        @file_options ||= I18n.t("database_admin.menu_order").each_with_object({}) do |file_option, h|
+          # special anonymous ftp sites
+          if file_option =~ %r{^ftps?://([^/]*)/}
```
carbonin: This might be easier to read without the regex. None of the other file options are uri formatted right? How about something like this:

```ruby
if %w(ftp ftps).include?(URI(file_option).scheme)
```

Then we can also avoid the global match variable by using the host in the URI.

lib/manageiq/appliance_console/database_admin.rb
kbrock: @kbrock When I was typing this I was thinking of modifying ask_with_menu to add that as an option

NickLaMuro: Maybe just make an issue for this?  I agree, that would definitely make this cleaner.  Maybe also have it as an option to opt of ( something like `ask_with_menu(*file_menu_args, :cancel_option => false)` maybe?).

...which reminds me... I really need to add an issue for the "UI stuff" I have wanted to refactor in this...

PR #66, opened 2018-09-25
Support for Swift Backup/Restore

jerryk55:
Ability to restore backups previous saved to Swift Object STorage.

In support of the RFE defined in BZ https://bugzilla.redhat.com/show_bug.cgi?id=1615488

Related PRs are https://github.com/ManageIQ/manageiq-ui-classic/pull/4684
https://github.com/ManageIQ/manageiq/pull/17967
https://github.com/ManageIQ/manageiq-gems-pending/pull/371
and
https://github.com/ManageIQ/manageiq-schema/pull/259.

@NickLaMuro @carbonin @roliveri please review.  Thanks.

### DIFF
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -36,7 +36,8 @@ module ManageIQ
 
       WARN
 
-      attr_reader :action, :backup_type, :task, :task_params, :delete_agree, :uri, :filename
+      attr_accessor :uri
+      attr_reader :action, :backup_type, :task, :task_params, :delete_agree, :filename
 
       def initialize(action = :restore, input = $stdin, output = $stdout)
         super(input, output)
@@ -172,35 +173,33 @@ module ManageIQ
           Example: 'openstack_user'
         PROMPT
 
-        @filename         = just_ask(*filename_prompt_args) unless action == :restore
-        @uri              = URI(ask_for_uri(*remote_file_prompt_args_for("swift")))
-        user              = just_ask(swift_user_prompt)
-        pass              = ask_for_password("password for #{user}")
-        region            = just_ask("OpenStack Swift Region")
-        @uri.port         = just_ask("OpenStack Swift Port", "5000")
-        security_protocol = ask_with_menu(*security_protocol_menu_args)
-        api_version       = ask_with_menu(*api_version_menu_args)
-        domain_ident      = just_ask("OpenStack V3 Domain Identifier") if api_version == "v3"
+        @filename  = just_ask(*filename_prompt_args) { |q| q.readline = false } unless action == :restore
+        @uri       = URI(ask_for_uri(*remote_file_prompt_args_for("swift")) { |q| q.readline = false })
+        @task      = "evm:db:#{action}:remote"
+        user       = just_ask(swift_user_prompt) { |q| q.readline = false }
+        pass       = ask_for_password("password for #{user}") { |q| q.readline = false }
+        @uri.query = swift_query_elements.join('&').presence
 
-        @task          = "evm:db:#{action}:remote"
-        query_elements = []
-        query_elements << "region=#{region}"                       if region.present?
-        query_elements << "api_version=#{api_version}"             if api_version.present?
-        query_elements << "domain_id=#{domain_ident}"              if domain_ident.present?
-        query_elements << "security_protocol=#{security_protocol}" if security_protocol.present?
-        @uri.query = query_elements.join('&').presence
-
-        params = [
-          "--",
-          {
-            :uri          => @uri.to_s,
-            :uri_username => user,
-            :uri_password => pass
-          }
-        ]
-
+        params = {
+          :uri          => @uri.to_s,
+          :uri_username => user,
+          :uri_password => pass
+        }
         @task        = "evm:db:#{action}:remote"
         @task_params = ["--", params]
+      end
+
+      def swift_query_elements
+        region            = just_ask("OpenStack Swift Region") { |q| q.readline = false }
+        @uri.port         = just_ask("OpenStack Swift Port", "5000") { |q| q.readline = false }
+        security_protocol = ask_with_menu(*security_protocol_menu_args)
+        api_version       = ask_with_menu(*api_version_menu_args) { |q| q.readline = false }
+        domain_ident      = just_ask("OpenStack V3 Domain Identifier") { |q| q.readline = false } if api_version == "v3"
+        query_elements    = []
+        query_elements    << "region=#{region}"                       if region.present?
+        query_elements    << "api_version=#{api_version}"             if api_version.present?
+        query_elements    << "domain_id=#{domain_ident}"              if domain_ident.present?
+        query_elements    << "security_protocol=#{security_protocol}" if security_protocol.present?
       end
 
       def ask_to_delete_backup_after_restore
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -50,6 +50,10 @@ module ManageIQ
         backup_type == "local".freeze
       end
 
+      def object_store_backup?
+        backup_type == "s3".freeze || backup_type == "swift".freeze
+      end
+
       def ask_questions
         setting_header
         say(DB_DUMP_WARNING) if action == :dump
@@ -328,24 +332,21 @@ module ManageIQ
       end
 
       def filename_prompt_args
-        default   = action == :dump ? DB_DEFAULT_DUMP_FILE : DB_RESTORE_FILE
-        validator = LOCAL_FILE_VALIDATOR if action == :restore && backup_type == LOCAL_FILE
-        prompt = "location of the local restore file" if action == :restore
-        if backup_type == S3_FILE || backup_type == SWIFT_FILE
-          prompt ||= object_store_file_prompt
+        return restore_prompt_args if action == :restore
+        default = action == :dump ? DB_DEFAULT_DUMP_FILE : DB_RESTORE_FILE
+        prompt  = "location to save the #{action} file to"
+        if object_store_backup?
+          prompt  = "name for the remote #{action} file"
           default = File.basename(default)
-        else
-          prompt ||= local_file_prompt
         end
+        [prompt, default, nil, "file that exists"]
+      end
+
+      def restore_prompt_args
+        default   = DB_RESTORE_FILE
+        validator = LOCAL_FILE_VALIDATOR if local_backup?
+        prompt    = "location of the local restore file"
         [prompt, default, validator, "file that exists"]
-      end
-
-      def local_file_prompt
-        "location to save the #{action} file to"
-      end
-
-      def object_store_file_prompt
-        "name for the remote #{action} file"
       end
 
       def remote_file_prompt_args_for(remote_type)
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -185,6 +185,7 @@ module ManageIQ
           :uri_username => user,
           :uri_password => pass
         }
+        params[:remote_file_name] = filename if filename
         @task        = "evm:db:#{action}:remote"
         @task_params = ["--", params]
       end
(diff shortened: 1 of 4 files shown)


lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -222,6 +276,15 @@ def file_menu_args
         ]
       end
 
+      def security_protocol_menu_args
+        [
+          "OpenStack Security Protocol",
+          SECURITY_PROTOCOL_OPTIONS,
```
carbonin: If this is the only place this is used, I think it would be better to just list the strings here, no need to declare a constant at the class level.

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -213,6 +258,15 @@ def file_options
         end
       end
 
+      def api_version_menu_args
+        [
+          "OpenStack API Version",
+          API_VERSION_OPTIONS,
```
carbonin: Same thing here, I don't think this needs to be a constant.

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -153,6 +168,36 @@ def ask_custom_file_options(server_uri)
...
+        user              = just_ask(USER_PROMPT)
+        pass              = ask_for_password("password for #{user}")
+        region            = just_ask("OpenStack Swift Region")
+        port              = just_ask("OpenStack Swift Port", "5000")
+        security_protocol = ask_with_menu(*security_protocol_menu_args)
+        api_version       = ask_with_menu(*api_version_menu_args)
+        domain_ident      = just_ask("OpenStack V3 Domain Identifier") if api_version == "v3"
+
+        @task          = "evm:db:#{action}:remote"
+        @uri           = URI.parse("#{URI(@uri).scheme}://#{URI(@uri).host}:#{port}#{URI(@uri).path}")
```
carbonin: What is happening here? I'm not sure what this line is trying to accomplish, but my gut feeling is that it can be done in a better way.

NickLaMuro:
@miq-bot remove_label unmergable

NickLaMuro:
\*grumbles\*

@miq-bot remove_label unmergeable

happy!?

lib/manageiq/appliance_console/database_admin.rb
NickLaMuro: To add to this, this is basically parsing the `@uri` 4 times, and it seems pretty unecessary.  I think doing the following would suffice:

```ruby
@uri              = URI(ask_for_uri(*remote_file_prompt_args_for("swift")))
# ...
@uri.port         = just_ask("OpenStack Swift Port", "5000")
```

And then you can remove this line we are commenting on entirely.

lib/manageiq/appliance_console/database_admin.rb
jerryk55: yup.

lib/manageiq/appliance_console/database_admin.rb
jerryk55: agreed.

lib/manageiq/appliance_console/database_admin.rb
jerryk55: Adding the port into the URI.  I'm sure there are *other* ways to do it.  I am open to suggestions other than this one.

jerryk55:
@NickLaMuro doh!  That works and is much easier.  Thanks!

lib/manageiq/appliance_console/database_admin.rb
jerryk55: I can do that.

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -153,6 +154,36 @@ def ask_custom_file_options(server_uri)
         @task_params = ["--", params]
       end
 
+      def ask_swift_file_options
+        require 'uri'
+        @uri              = ask_for_uri(*remote_file_prompt_args_for("swift"))
+        user              = just_ask(USER_PROMPT)
```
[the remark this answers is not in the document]
jerryk55: Yup.  good catch...

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -213,6 +244,15 @@ def file_options
         end
       end
 
+      def api_version_menu_args
+        [
+          "OpenStack API Version",
+          [["Keystone v2".freeze, "v2".freeze], ["Keystone v3".freeze, "v3".freeze], [CANCEL, nil]].freeze,
```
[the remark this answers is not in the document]
jerryk55: None it is.

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -222,6 +262,15 @@ def file_menu_args
         ]
       end
 
+      def security_protocol_menu_args
+        [
+          "OpenStack Security Protocol",
+          [["SSL without validation".freeze, "ssl".freeze], ["SSL".freeze, "ssl-with-validation".freeze], ["Non-SSL".freeze, "non-ssl".freeze], [CANCEL, nil]].freeze,
```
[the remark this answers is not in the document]
jerryk55: Yup.  "None" again.

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -153,6 +154,35 @@ def ask_custom_file_options(server_uri)
         @task_params = ["--", params]
       end
 
+      def ask_swift_file_options
+        require 'uri'
+        @uri              = URI(ask_for_uri(*remote_file_prompt_args_for("swift")))
```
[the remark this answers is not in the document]
jerryk55: Got it.

jerryk55:
Note that I am still working the issues with the spec tests.  I am aware of the failures.

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -11,6 +11,7 @@ class DatabaseAdmin < HighLine
       SMB_FILE   = "smb".freeze
       S3_FILE    = "s3".freeze
       FTP_FILE   = "ftp".freeze
+      SWIFT_FILE = "swift".freeze
```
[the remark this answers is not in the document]
kbrock: yea, these have been removed / are not necessary anymore. we are now just hardcoding them a few times in the specs

jerryk55:
@NickLaMuro @carbonin specs pushed out.  Not aware of any other outstanding comments but I'm open to any others - If we can get this pushed out for the Sprint that would be awesome.

spec/database_admin_spec.rb
```diff
@@ -393,6 +401,174 @@
...
+      let(:filename)             { File.basename(example_uri) }
+      let(:user)                 { 'foobar' }
+      let(:pass)                 { 'supersecret' }
+      let(:uri_prompt)           { "Enter the location of the remote backup file\nExample: #{example_uri}: " }
+      let(:user_prompt)          { "\n?  Enter the User Name with access to this file.\nExample: 'openstack_user': " }
+      let(:pass_prompt)          { "Enter the password for #{user}" }
+      let(:region_prompt)        { "Enter the OpenStack Swift Region: " }
+      let(:port_prompt)          { "Enter the OpenStack Swift Port: |5000| " }
+      let(:domain_prompt)        { "OpenStack V3 Domain Identifier" }
+      let(:security_protocol_prompt) { "OpenStack Security Protocol\n\n1) SSL without validation\n2) SSL\n3) Non-SSL\n4) None\n\nChoose the openstack security protocol: |3| " }
```
NickLaMuro: _**Pedantic**_ (No need to change):  If you want to make this align without having to move the rest of the `let` statements over, you could change this variable name to `:ssl_prompt` 😉 

(_@NickLaMuro really just felt bad asking to align the the brackets and have the values go further over the 80 char limit without providing a suggestion_)

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -260,16 +329,23 @@ def should_split_output?
 
       def filename_prompt_args
         default   = action == :dump ? DB_DEFAULT_DUMP_FILE : DB_RESTORE_FILE
-        validator = LOCAL_FILE_VALIDATOR if action == :restore && local_backup?
-        [local_file_prompt, default, validator, "file that exists"]
+        validator = LOCAL_FILE_VALIDATOR if action == :restore && backup_type == LOCAL_FILE
+        prompt = "location of the local restore file" if action == :restore
+        if backup_type == S3_FILE || backup_type == SWIFT_FILE
```
NickLaMuro: These constants no longer exist, and can't be used.

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -260,16 +329,23 @@ def should_split_output?
 
       def filename_prompt_args
         default   = action == :dump ? DB_DEFAULT_DUMP_FILE : DB_RESTORE_FILE
-        validator = LOCAL_FILE_VALIDATOR if action == :restore && local_backup?
-        [local_file_prompt, default, validator, "file that exists"]
+        validator = LOCAL_FILE_VALIDATOR if action == :restore && backup_type == LOCAL_FILE
```
NickLaMuro: This should be:

```ruby
validator = LOCAL_FILE_VALIDATOR if action == :restore && local_backup?
```

Since `DatabaseAdmin#local_backup?` was added in https://github.com/ManageIQ/manageiq-appliance_console/pull/69 as a result of me asking @kbrock to remove the constants that were no longer being used (since he switched to the menu being generated from a `I18n` file).

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -260,16 +329,23 @@ def should_split_output?
...
+        [prompt, default, validator, "file that exists"]
       end
 
       def local_file_prompt
-        if action == :restore
-          "location of the local restore file"
-        else
-          "location to save the #{action} file to"
-        end
+        "location to save the #{action} file to"
```
NickLaMuro: I get adding the change for the below, but this doesn't make sense for `:restore` actions for local files (as this question only gets hit for `:restore` actions for local files.  All other provider types are doing the following (even pulled from your swift changes above):

```ruby
@filename  = just_ask(*filename_prompt_args) { |q| q.readline = false } unless action == :restore
```

Except the `ask_local_file_options`, in which this question makes perfect sense since the rake tasks args for `evm:db:restore:local` differ from `evm:db:restore:remote`.

NickLaMuro: This got addressed.  Resolving.

lib/manageiq/appliance_console/database_admin.rb
NickLaMuro: This was addressed.  Resolving.

lib/manageiq/appliance_console/database_admin.rb
NickLaMuro: Addressed.  Resolving.

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -6,6 +6,19 @@ module ApplianceConsole
...
+      V3_VERSION          = "v3".freeze
+      API_VERSION_OPTIONS = [[V2_VERSION_PROMPT, V2_VERSION], [V3_VERSION_PROMPT, V3_VERSION], [CANCEL, nil]].freeze
+
+      SSL_WO_VALIDATION_PROMPT   = "SSL without validation".freeze
+      SSL_WO_VALIDATION          = "ssl".freeze
+      SSL_WITH_VALIDATION_PROMPT = "SSL".freeze
+      SSL_WITH_VALIDATION        = "ssl-with-validation".freeze
+      NON_SSL_PROMPT             = "Non-SSL".freeze
+      NON_SSL                    = "non-ssl".freeze
+      SECURITY_PROTOCOL_OPTIONS  = [[SSL_WO_VALIDATION_PROMPT, SSL_WO_VALIDATION], [SSL_WITH_VALIDATION_PROMPT, SSL_WITH_VALIDATION], [NON_SSL_PROMPT, NON_SSL], [CANCEL, nil]].freeze
```
carbonin: I don't see any of these constants used anywhere in this patch. Can we remove them?

NickLaMuro: Oh... good catch.  Were they meant for `api_version_menu_args` and `security_protocol_menu_args`?

If we go back to using these instead, just make sure you change `CANCEL` to `"None"`.

PR #69, opened 2018-10-04
Anonymous FTP request single file prompt (vs 2)

kbrock:
We were prompting for a local data filename and a remote url.
But since this is streaming, it does not make sense.

This now:

- properly populates the `:remote_file_name value`
- hides/shows the prompt selectively for dump, backup, restore
- has configurable rake options

related to:
- https://bugzilla.redhat.com/show_bug.cgi?id=1632433
- ManageIQ/manageiq-appliance_console#65

configuration has been tweaked:

```yaml
  database_admin:
    menu_order:
    - local
    - ftp://ftp.example.com/incoming/
    local: Local file
    prompts:
      ftp.example.com:
        enabled_for: ["backup", "dump"]
        filename_text: "The case number dash (-) filename. (e.g.: 12345-db.backup)"
        filename_validator: "^[0-9]{4,}-..*"
        rake_options:
          skip_directory: true
```

### DIFF
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -205,7 +205,7 @@ module ManageIQ
           # special anonymous ftp sites are defined by uri
           uri = URI(file_option)
           if uri.scheme
-            h["#{uri.scheme} to #{uri.host}"] = file_option
+            h["#{uri.scheme} to #{uri.host}"] = file_option unless skip_file_location?(uri.host)
           else
             h[I18n.t("database_admin.#{file_option}")] = file_option
           end
@@ -228,12 +228,22 @@ module ManageIQ
       private
 
       def ask_custom_file_prompt(hostname)
-        # hostname has a period in it, so we need to look it up by [] instead of the traditional i18n method
-        prompts = I18n.t("database_admin.prompts", :default => {})[hostname.to_sym]
+        prompts = custom_endpoint_config_for(hostname)
         prompt_text  = prompts && prompts[:filename_text] || "Target filename for backup".freeze
         prompt_regex = prompts && prompts[:filename_validator]
         validator    = prompt_regex ? ->(x) { x.to_s =~ /#{prompt_regex}/ } : ->(x) { x.to_s.present? }
         just_ask(prompt_text, nil, validator)
+      end
+
+      def skip_file_location?(hostname)
+        config = custom_endpoint_config_for(hostname)
+        return false unless config && config[:enabled_for].present?
+        !Array(config[:enabled_for]).include?(action.to_s)
+      end
+
+      def custom_endpoint_config_for(hostname)
+        # hostname has a period in it, so we need to look it up by [] instead of the traditional i18n method
+        I18n.t("database_admin.prompts", :default => {})[hostname.to_sym]
       end
 
       def should_exclude_tables?
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -6,12 +6,6 @@ module ManageIQ
     class DatabaseAdmin < HighLine
       include ManageIQ::ApplianceConsole::Prompts
 
-      LOCAL_FILE = "local".freeze
-      NFS_FILE   = "nfs".freeze
-      SMB_FILE   = "smb".freeze
-      S3_FILE    = "s3".freeze
-      FTP_FILE   = "ftp".freeze
-
       DB_RESTORE_FILE      = "/tmp/evm_db.backup".freeze
       DB_DEFAULT_DUMP_FILE = "/tmp/evm_db.dump".freeze
       LOCAL_FILE_VALIDATOR = ->(a) { File.exist?(a) }.freeze
@@ -36,6 +30,10 @@ module ManageIQ
 
         @action      = action
         @task_params = []
+      end
+
+      def local_backup?
+        backup_type == "local".freeze
       end
 
       def ask_questions
@@ -155,7 +153,7 @@ module ManageIQ
       end
 
       def ask_to_delete_backup_after_restore
-        if action == :restore && backup_type == LOCAL_FILE
+        if action == :restore && local_backup?
           say("The local database restore file is located at: '#{uri}'.\n")
           @delete_agree = agree("Should this file be deleted after completing the restore? (Y/N): ")
         end
@@ -218,7 +216,7 @@ module ManageIQ
         [
           action == :restore ? "Restore Database File" : "#{action.capitalize} Output File Name",
           file_options,
-          LOCAL_FILE,
+          "local",
           nil
         ]
       end
@@ -252,7 +250,7 @@ module ManageIQ
 
       def filename_prompt_args
         default   = action == :dump ? DB_DEFAULT_DUMP_FILE : DB_RESTORE_FILE
-        validator = LOCAL_FILE_VALIDATOR if action == :restore && backup_type == LOCAL_FILE
+        validator = LOCAL_FILE_VALIDATOR if action == :restore && local_backup?
         [local_file_prompt, default, validator, "file that exists"]
       end
 
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -142,11 +142,11 @@ module ManageIQ
         @filename = ask_custom_file_prompt(hostname)
         @uri      = server_uri
 
-        params = {
-          :uri              => uri,
-          :skip_directory   => true,
-          :remote_file_name => filename
-        }
+        params = {:uri => uri, :remote_file_name => filename}
+
+        if (custom_params = custom_endpoint_config_for(hostname))
+          params.merge!(custom_params[:rake_options]) if custom_params[:rake_options]
+        end
 
         @task        = "evm:db:#{action}:remote"
         @task_params = ["--", params]
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -140,14 +140,15 @@ module ManageIQ
       end
 
       def ask_custom_file_options(server_uri)
-        @filename    = just_ask(*filename_prompt_args) unless action == :restore
-        sample_case  = server_uri.split("/").last
-        hostname     = URI(server_uri).host
-        uri_filename = ask_custom_prompt(hostname, 'filename', "Target filename (e.g.: #{sample_case})")
-        @uri         = server_uri.gsub(sample_case, uri_filename)
+        hostname  = URI(server_uri).host
+        @filename = ask_custom_prompt(hostname, 'filename', "Target filename for backup")
+        @uri      = server_uri
 
-        params = { :uri => uri }
-        params[:remote_file_name] = filename if filename
+        params = {
+          :uri              => uri,
+          :skip_directory   => true,
+          :remote_file_name => filename
+        }
 
         @task        = "evm:db:#{action}:remote"
         @task_params = ["--", params]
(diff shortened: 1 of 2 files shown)


lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -140,14 +140,12 @@ def ask_ftp_file_options
       end
 
       def ask_custom_file_options(server_uri)
-        @filename    = just_ask(*filename_prompt_args) unless action == :restore
-        sample_case  = server_uri.split("/").last
-        hostname     = URI(server_uri).host
-        uri_filename = ask_custom_prompt(hostname, 'filename', "Target filename (e.g.: #{sample_case})")
-        @uri         = server_uri.gsub(sample_case, uri_filename)
+        hostname  = URI(server_uri).host
+        @filename = ask_custom_prompt(hostname, 'filename', "Target filename for backup")
```
NickLaMuro: I realize this was added in the previous PR, and I wasn't really the reviewer, but why are we "passing" the default prompt if this method is only used in one spot?

I was reviewing a similar PR that @carbonin was (where he asked to axe some constants), but this seems like it makes more sense to just have this as a constant, or a frozen string in the method itself instead of passing it from here.

NickLaMuro: Also (not part of this line of code), but I think you can delete the constants at the top of this file for `LOCAL_FILE`, `NFS_FILE`, etc. since we don't use them any more (as of #65).

Separate commit/PR maybe?  I think you would have a lot of tests that would need to be updated, for what it is worth.

spec/database_admin_spec.rb
```diff
@@ -1736,38 +1736,32 @@ def confirm_and_execute
     end
 
     describe "#ask_custom_file_options" do
```
NickLaMuro: Looking back at `describe "#ask_custom_file_options" do` in the last PR, you only added this for the current context this is under (I think just "DB Dump").

Can you add specs for the other two?  (you should be able to grep for the above to find the other places where this is at).

lib/manageiq/appliance_console/database_admin.rb
kbrock: ok, will make the prompt less generic.

Those constants are used all over the specs :(  They have been simplified quite a bit (now just have `'ftp'` vs the full text string) -- we can hardcode phrases like "ftp" in the specs.

kbrock:
ok

- added the missing tests (they felt a little copy paste),
- fixed custom_file prompt (removed the genericness)
- removed the constants (that were only used in specs)

keep 'em comin'

kbrock:
@NickLaMuro What I used to configure the appliance console in local development to act like the appliance vm and pickup my localization changes

```
$ pwd
/Users/kbrock/src/manageiq-appliance_console
$ cat ~/src/manageiq/productization/appliance_console/locales/appliance/en_productization.yml
---
en:
  database_admin:
    menu_order:
    - local
    - ftp://ftp.example.com/incoming
    local: Local file
    prompts:
      ftp.example.com:
        filename_text: "The case number dash (-) filename. (e.g.: 12345-db.backup)"
        filename_validator: "^[0-9]{4,}-..*"
$ bundle exec irb -I lib
irb> module ManageIQ; module ApplianceConsole; RAILS_ROOT = File.expand_path("../manageiq", Dir.pwd); end; end
# => "/Users/kbrock/src/manageiq"  
irb> require 'manageiq/appliance_console/i18n'
# => true  
irb> I18n.t("database_admin.prompts", :default => {})
# => {
  :"ftp.example.com"=>{
    :filename_text=>"The case number dash (-) filename. (e.g.: 12345-db.backup)",
    :filename_validator=>"^[0-9]{4,}-..*"
  }
}  
irb> I18n.load_path
# => [
  "/Users/kbrock/src/manageiq-appliance_console/locales/appliance/en.yml",
  "/Users/kbrock/src/manageiq/productization/appliance_console/locales/appliance/en_productization.yml"
]  
```

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -140,21 +138,22 @@ def ask_ftp_file_options
...
-        @uri         = server_uri.gsub(sample_case, uri_filename)
+        hostname  = URI(server_uri).host
+        @filename = ask_custom_file_prompt(hostname)
+        @uri      = server_uri
 
-        params = { :uri => uri }
-        params[:remote_file_name] = filename if filename
+        params = {
+          :uri              => uri,
+          :skip_directory   => true,
```
NickLaMuro: I am fine with this for now, but I think in the end, this should be part of the configuration, and not hard coded in.

NickLaMuro: Along with this, I really feel we should be hiding this option, or allowing it to be configured to do so, for the restore options since this is ultimately being implemented for an endpoint you won't be able to restore from.

Might need a follow up, but just my two cents.

kbrock:
WIPping to find a more configurable way of adding flags for specific backup options

kbrock:
database_admin:
    menu_order:
    - local
    - ftp://ftp.example.com/incoming/
    local: Local file
    prompts:
      ftp.example.com:
        filename_text: "The case number dash (-) filename. (e.g.: 12345-db.backup)"
        filename_validator: "^[0-9]{4,}-..*"

kbrock:
> Along with this, I really feel we should be hiding this option, or allowing it to be configured to do so, for the restore options since this is ultimately being implemented for an endpoint you won't be able to restore from.
> -- @NickLaMuro 

Further clarifying this comment:

For our current use case, this option should be displayed for `backup` and `dump`, but should be hidden for the `restore` menu.
There are 2 possible solutions so far for configuration:

### the dry solution
@NickLaMuro suggests dry. It is no fun to have duplicate lists - duplication tends to go wrong quickly.

We display `menu_order` for all entries, unless an entry has `enable_for` key and the key does not include the current menu name (i.e.: dump, backup, restore).

```yaml
---
  database_admin:
    menu_order:
    - local
    - ftp://ftp.example.com/incoming/
    local: Local file
    prompts:
      ftp.example.com:
        enable_for: [dump, backup]
        filename_text: "The case number dash (-) filename. (e.g.: 12345-db.backup)"
        filename_validator: "^[0-9]{4,}-.+"
```

### the less logic solution
@kbrock suggests duplicating but removing special parameters that negate the list. Duplication stinks, but conditional logic is worse.

If `{restore, backup, dump}_menu_order` is available, display that. Otherwise fall back to display `menu_order`.

```yaml
---
  database_admin:
    menu_order:
    - local
    - ftp://ftp.example.com/incoming/
    restore_menu_order:
    - local
    - ftp://ftp.example.com/incoming/
    local: Local file
    prompts:
      ftp.example.com:
        filename_text: "The case number dash (-) filename. (e.g.: 12345-db.backup)"
        filename_validator: "^[0-9]{4,}-.+"
```

kbrock:
added rake_options and made custom endpoints enabled for various options (backup, dump, restore)

Does this work for you now @NickLaMuro ?

lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -140,21 +138,22 @@ def ask_ftp_file_options
       end
 
       def ask_custom_file_options(server_uri)
-        @filename    = just_ask(*filename_prompt_args) unless action == :restore
-        sample_case  = server_uri.split("/").last
-        hostname     = URI(server_uri).host
-        uri_filename = ask_custom_prompt(hostname, 'filename', "Target filename (e.g.: #{sample_case})")
-        @uri         = server_uri.gsub(sample_case, uri_filename)
+        hostname  = URI(server_uri).host
```
NickLaMuro: This is minor:

I found this a bit confusing to me for a bit since I saw you parsing out the `host` from the `server_uri`, and thought that was being being assigned to `@uri`.  Turns out that wasn't the case, but we also aren't using the `hostname` in any meaningful fashion except as a key.  Seems like we could just as easily do this:

```yaml
  database_admin:
    menu_order:
    - local
    - ftp://ftp.example.com/incoming/
    local: Local file
    prompts:
      ftp://ftp.example.com/incoming/:
        enabled_for: ["backup", "dump"]
        ...
```

And achieve the same affect by just passing around `server_uri` as the key.

* * * 

Again, this is minor, just thought it was a bit confusing from both a code and configuration perspective since there is two different "keys" being passed around.

spec/database_admin_spec.rb
```diff
@@ -752,48 +954,103 @@ def confirm_and_execute
...
+            3) Cancel
+
+            Choose the backup output file name: |1|
+          PROMPT
+        end
+
+        it "displays custom ftp option with enabled array" do
+          expect(I18n).to receive(:t).with("database_admin.menu_order").and_return(%w(local ftp://example.com/inbox/filename.txt))
+          expect(I18n).to receive(:t).with("database_admin.local").and_return("The Local file")
+          expect(I18n).to receive(:t).with("database_admin.prompts", :default => {}).at_least(:once).and_return(:"example.com" => { :enabled_for => %w(restore backup) })
```
NickLaMuro: _**Pedantic**_:  Wouldn't this be a known quantity of `.once` or `.twice`?  Would save some characters and would be more explicit with those since this shouldn't be a variable number of calls as far as I can tell.

spec/database_admin_spec.rb
```diff
@@ -752,48 +954,103 @@ def confirm_and_execute
...
+            2) ftp to example.com
+            3) Cancel
+
+            Choose the backup output file name: |1|
+          PROMPT
+        end
+
+        it "displays custom ftp option with enabled array" do
+          expect(I18n).to receive(:t).with("database_admin.menu_order").and_return(%w(local ftp://example.com/inbox/filename.txt))
+          expect(I18n).to receive(:t).with("database_admin.local").and_return("The Local file")
```
NickLaMuro: _**Pedantic**_:  These above two seem like they could be tossed in a `before` hook to save them being defined twice.

spec/database_admin_spec.rb
```diff
@@ -752,48 +954,103 @@ def confirm_and_execute
...
+
+            1) The Local file
+            2) ftp to example.com
+            3) Cancel
+
+            Choose the backup output file name: |1|
+          PROMPT
+        end
+
+        it "hids custom ftp option with disabled string" do
```
NickLaMuro: Typo:  `s/hids/hides`

spec/database_admin_spec.rb
NickLaMuro: Also, since this line is pretty long, could the prompt options be saved in a `let(:prompt_options)` to allow this to be more visable?

spec/database_admin_spec.rb
```diff
@@ -67,48 +67,175 @@
...
+        expect(subject.backup_type).to eq("ftp")
       end
 
       it "cancels when CANCEL option is choosen" do
         say "6"
         expect { subject.ask_file_location }.to raise_error signal_error
       end
+
+      # this is the complete implementation. the other 2 are paired down version of this
+      context "with localized file upload" do
```
NickLaMuro: _**Pedantic**_:  maybe "with I18n custom menu configs"?

If changed, make sure to change in the other two spots.

lib/manageiq/appliance_console/database_admin.rb
kbrock: Lets keep this as the host for now.
Using the full path would simplify the code and may be more clear
But I like reading the short prompt version

spec/database_admin_spec.rb
kbrock: The actual value for `enabled_for` are different for most tests. I'll make a helper for this.

I'll make the `twice` more explicit

lib/manageiq/appliance_console/database_admin.rb
NickLaMuro: Works for me.

If the above wasn't clear to on lookers, my above critique is pretty subjective, and I don't hold a hard opinion on this.

kbrock:
@miq-bot tag hammer/yes

carbonin:
@kbrock The console is actually a released gem so at some point when the backup/dump work stabilizes, we'll cut a release and include that in the hammer branch of manageiq-appliance

PR #72, opened 2018-11-13
Fix Database Menu Prompts

jerryk55:
The menus for the Database backup, dump, and restore should be changed.
They infer that they are prompting for the file *name* when in fact they are prompting for the type of the storage location on which the DB will be stored or from which it will be retrieved.  The easiest and shortest change is to change "Name" to "Type", and to add "Type" where neither is present.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1648986

@NickLaMuro @carbonin @roliveri please review.

### DIFF
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -265,7 +265,7 @@ module ManageIQ
 
       def file_menu_args
         [
-          action == :restore ? "Restore Database File" : "#{action.capitalize} Output File Name",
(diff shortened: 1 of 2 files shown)


carbonin:
How about "file source" for restore and "file destination" for everything else? Calling FTP the file "type" feels just as wrong.

jerryk55:
Sure thing!!

NickLaMuro:
I am mostly good with this.  Going to check when this got added (by me, and when really), but I probably will still 100% agree with this change.  Just want to make sure I understand why it was called this in the first place (if there was a reason).

PR #73, opened 2018-11-28
Use Optimist

NickLaMuro:
`Trollop` is a deprecated gem, and has been renamed/replaced by [`Optimist`](https://github.com/ManageIQ/optimist).

This change might be dependent on `manageiq` making changes to this as well.  I don't think this should be merged until we have fully released `hammer`, since we might be a chance we need to release this gem for a bug fix yet.  Just getting this and others prepared currently, and will un-`[WIP]` when I think it is safe to merge this one.

Though, not making a release with this change might make it so nightly builds might have trouble once https://github.com/ManageIQ/manageiq/pull/18246 is merged, since that might not contain the dependency needed for the `cli` to run.  Will have to look into how this gem is installed on an appliance...

Update:  Installed via https://github.com/ManageIQ/manageiq-appliance/blob/1f8a6dcc/manageiq-appliance-dependencies.rb#L2 so this shouldn't break nightlies by merging the the above without this one, there will just be both `trollop` and `optimist` on an appliance for a while, along with the warning (since we don't have the version of `trollop` pegged in the gemspec.).

Links
-----
* Org search of current usage of old gem (hides some irrelevant repos):  [search](https://github.com/search?l=&q=Trollop+repo%3AManageIQ%2Fmanageiq+repo%3AManageIQ%2Fmanageiq-release+repo%3AManageIQ%2Fmanageiq-appliance_console+repo%3AManageIQ%2Fmanageiq-appliance-build+repo%3AManageIQ%2Fmanageiq-api+repo%3AManageIQ%2Fmanageiq-providers-vmware+repo%3AManageIQ%2Fhttpd_configmap_generator+repo%3AManageIQ%2Fguides&type=Code)

### DIFF
--- a/lib/manageiq/appliance_console/cli.rb
+++ b/lib/manageiq/appliance_console/cli.rb
@@ -1,4 +1,4 @@
-require 'trollop'
+require 'optimist'
 require 'pathname'
 
 # support for appliance_console methods
@@ -106,7 +106,7 @@ module ApplianceConsole
 
     def parse(args)
       args.shift if args.first == "--" # Handle when called through script/runner
-      self.options = Trollop.options(args) do
+      self.options = Optimist.options(args) do
         banner "Usage: appliance_console_cli [options]"
 
         opt :host,     "/etc/hosts name",    :type => :string,  :short => 'H'
@@ -147,7 +147,7 @@ module ApplianceConsole
         opt :extauth_opts,         "External Authentication Options",   :type => :string
         opt :server,               "{start|stop|restart} actions on evmserverd Server",   :type => :string
       end
-      Trollop.die :region, "needed when setting up a local database" if region_number_required? && options[:region].nil?
+      Optimist.die :region, "needed when setting up a local database" if region_number_required? && options[:region].nil?
       self
     end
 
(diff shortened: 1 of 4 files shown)


kbrock:
Heh - I remember writing that line that is blowing up your code climate.
luckily it git blames to someone else so... :running_man:

Maybe rebase this? Is it really a `WIP`?

NickLaMuro:
> Is it really a WIP?

@kbrock read description.

Fryguy:
cc @simaishi

NickLaMuro:
@Fryguy @kbrock updated with the latest master, and removed the `[WIP]` from the PR.

I was originally thinking that we could wait to merge until after https://github.com/ManageIQ/manageiq-appliance_console/pull/74 since that fixed some `hakiri` warnings we were also seeing, but maybe we aren't using that any more?

Do as you wish.  Thanks.

kbrock:
sigh. I've found it is very tricky to keep "complexity" down on cli classes.
Lots of duplicate load parameter, check parameter

which explains why there are so many cli libraries.
and in the end, they are kinda all the same.
And rubocop / codeclimate / mutant all complain about duplicate code or lots of `if` checks

I'm still :+1: on this PR

PR #74, opened 2018-12-03
Fix hakiri warnings on `i18n`

NickLaMuro:
By using `manageiq` versions of rails dependencies.  This PR does two things:

- use `rails-i18n` instead of `i18n`, matching `manageiq` version
- matches the versions of `activesupport`/`activerecord` to `manageiq`

Version matching done at for the current versions at the time of writing:

https://github.com/ManageIQ/manageiq/blob/4cdafdbf/Gemfile#L59-L60

This ideally will avoid multiple versions of the gems on the system, as well as removing the `hakiri` warning for `i18n`.

**Note**:  We probably should wait to merge this until after the `hammer` release, just to ease bug fixes leading up to it.  Will toss a `[WIP]` on this for the time being as a reminder of that.

### DIFF
--- a/manageiq-appliance_console.gemspec
+++ b/manageiq-appliance_console.gemspec
@@ -20,15 +20,15 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_runtime_dependency "activerecord",            ">= 4.2.2"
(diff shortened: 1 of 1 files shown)


NickLaMuro:
@Fryguy Well, `hammer` has now been released, but I also don't see `hakiri` in the list of checks... so maybe this doesn't have a purpose?

Anyway, removed the `[WIP]`, so do as you wish.

Fryguy:
I am good with this, but it conflicted with #73 , so can you please rebase?

PR #76, opened 2018-12-12
Fix logging when configure db

yrudman:
**Issue**: 
  last operator in `say_error`is throwing error and execution never reached `log_error`

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1625377

@miq-bot add-label  bug

### DIFF
--- a/lib/manageiq/appliance_console/database_configuration.rb
+++ b/lib/manageiq/appliance_console/database_configuration.rb
@@ -217,8 +217,8 @@ FRIENDLY
     def validated
       !!validate!
     rescue => err
-      say_error(__method__, err.message)
       log_error(__method__, err.message)
+      say_error(__method__, err.message)
       false
     end
 


yrudman:
@miq-bot add_label bug

lib/manageiq/appliance_console/database_configuration.rb
```diff
@@ -217,8 +217,8 @@ def self.region
     def validated
       !!validate!
     rescue => err
-      say_error(__method__, err.message)
       log_error(__method__, err.message)
+      say_error(__method__, err.message)
       false
```
carbonin: This line will never get run. Can we remove it?

carbonin:
We can merge this for now, but it looks like the `validated` method is just broken right now.

It will never return false and I can't tell what that might be breaking. Can you open up a follow up to fix that method?

PR #82, opened 2019-02-19
Update repmgrd service name for postgres 10

carbonin:


carbonin:
I'm going to un-wip this one since this is a released gem. We can just merge this one then do the gem release/update dance when we're ready.

PR #83, opened 2019-03-01
Handle existing certs and support retun of cert generation

jvlcek:
This PR will address:
   https://bugzilla.redhat.com/show_bug.cgi?id=1670138

This PR addresses a couple of related issues.

1 - If the self signed certs, that are auto-generated the first time the appliance is started,
    already exists "getcert request" would not overwrite them. This PR addresses this by first
    moving them.

2 - While working to resolve the above issue I realized that if "getcert request" was already
    successfully and certs are being track, then re-running the "getcert request" is actually
    a no-op, unless tracking is stopped with "getcert stop-tracking"
    This PR addresses this by running "getcert stop-tracking" when certs are already being tracked

    NOTE: If the certs are still valid on the cert authority the same certs will be provided.

**Testing**
One way to test this is to use the appliance_console to request a new ipa cert after the appliance
has been started at least once, which will create self signed certs if certs are not found.

The self signed certs will be created here:
```
/var/www/miq/vmdb/certs/server.cer
/var/www/miq/vmdb/certs/server.cer.key
```

```bash
# Start the appliance
systemctl start evmserverd

# Copy the self signed certs to x.SELF_SIGNED
cp /var/www/miq/vmdb/certs/server.cer /var/www/miq/vmdb/certs/server.cer.SELF_SIGNED
cp /var/www/miq/vmdb/certs/server.cer.key /var/www/miq/vmdb/certs/server.cer.key.SELF_SIGNED

# Run the appliance_console for appliance_console_cli to configure external auth with IPA, e.g:
appliance_console_cli --ipaserver=<my IPA server fully qualified hostname> --ipaprincipal=admin --ipapassword=<my IPA server's admin password>

# Run the appliance_console_cli to request signed certs
appliance_console_cli        --ca=ipa        --http-cert

# Ensure the certs created just created by the above command are different from the saved .SEFL_SIGNED ones.
diff -q /var/www/miq/vmdb/certs/server.cer /var/www/miq/vmdb/certs/server.cer.SELF_SIGNED
Files server.cer and server.cer.SELF_SIGNED differ

diff -q /var/www/miq/vmdb/certs/server.cer.key /var/www/miq/vmdb/certs/server.cer.key.SELF_SIGNED
Files server.cer.key and server.cer.key.SELF_SIGNED differ
```

### DIFF
--- a/lib/manageiq/appliance_console/certificate.rb
+++ b/lib/manageiq/appliance_console/certificate.rb
@@ -1,5 +1,6 @@
 require "awesome_spawn"
-require 'linux_admin'
+require "linux_admin"
+require "fileutils"
 
 module ManageIQ
 module ApplianceConsole
@@ -40,8 +41,11 @@ module ApplianceConsole
     end
 
     def request
+      undo_tracking if complete?
+
       if should_request_key?
         principal.register
+        remove_key_pair
         request_certificate
         # NOTE: status probably changed
         set_owner_of_key unless rejected?
@@ -110,6 +114,22 @@ module ApplianceConsole
     end
 
     private
+
+    def remove_key_pair
+      FileUtils.rm_f(cert_filename) if File.exist?(cert_filename)
+      FileUtils.rm_f(key_filename)  if File.exist?(key_filename)
+    end
+
+    def undo_tracking
+      stop_tracking
+      FileUtils.rm_f(root_filename) if File.exist?(root_filename)
+      remove_key_pair
+      clear_status
+    end
+
+    def stop_tracking
+      AwesomeSpawn.run!("/usr/bin/getcert", :params => ["stop-tracking", "-f", cert_filename, "-k", key_filename])
+    end
 
     def request_first
       params = {
--- a/spec/certificate_authority_spec.rb
+++ b/spec/certificate_authority_spec.rb
@@ -40,7 +40,7 @@ describe ManageIQ::ApplianceConsole::CertificateAuthority do
 
     it "should configure http" do
       ipa_configured(true)
-      expect_run(/getcert/, anything, response) # getcert returns: the certificate already exist
+      expect_run(/getcert/, anything, response).at_least(3).times
 
       expect(LinuxAdmin::Service).to receive(:new).and_return(double("Service", :restart => true))
       expect(LinuxAdmin::Service).to receive(:new).and_return(double(:enable => double(:start => nil)))
@@ -66,7 +66,7 @@ describe ManageIQ::ApplianceConsole::CertificateAuthority do
 
     it "should configure postgres client" do
       ipa_configured(true)
-      expect_run(/getcert/, anything, response) # getcert returns: the certificate already exist
+      expect_run(/getcert/, anything, response).at_least(3).times
 
       allow(File).to receive(:exist?).and_return(true)
       expect(LinuxAdmin::Service).to receive(:new).and_return(double(:enable => double(:start => nil)))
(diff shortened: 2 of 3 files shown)


jvlcek:
@miq-bot add_label bug

jvlcek:
@kbrock Please review

jvlcek:
@miq-bot assign @carbonin

jvlcek:
@carbonin and @kbrock Please review.
Can I get a `pass` on the codecliamte issue of one too many methods in a class?

carbonin:
@jvlcek can you remove the `Fixes` line from your commit? This needs to be included in a versioned release of the console gem and added to the appliance before the bug can be moved to POST

lib/manageiq/appliance_console/certificate.rb
```diff
@@ -40,8 +41,11 @@ def initialize(options = {})
     end
 
     def request
+      undo_tracking if complete?
```
carbonin: Be doing this first wouldn't we remove the self signed certs before saving them off? Or would we never get into the `if should_request_key?` block if we are `complete?` here?

lib/manageiq/appliance_console/certificate.rb
```diff
@@ -40,8 +41,11 @@ def initialize(options = {})
...
+
       if should_request_key?
         principal.register
+        save_self_signed
```
carbonin: Is this really specific to self signed certs or will it just move whatever is at the `key_filename`/`cert_filename` location?

For example, if I run this after having retrieved the certs from the authority, I wouldn't be saving self-signed certs, right? They would be the real ones. Or am I misunderstanding?

lib/manageiq/appliance_console/certificate.rb
```diff
@@ -111,6 +115,23 @@ def enable_certmonger
...
     private
 
+    def save_self_signed
+      FileUtils.mv(key_filename, "#{key_filename}.ORIG", :force => true)  unless File.exist?("#{key_filename}.ORIG")
+      FileUtils.mv(cert_filename, "#{cert_filename}.ORIG", :force => true)  unless File.exist?("#{cert_filename}.ORIG")
+    end
+
+    def undo_tracking
+      stop_tracking
+      FileUtils.rm_f(root_filename) if File.exist?(root_filename)
```
carbonin: Should this file be saved in `save_self_signed` as well?

lib/manageiq/appliance_console/certificate.rb
jvlcek: Thank you for the review input @carbonin.

This line: `undo_tracking if complete?` will only invoke `undo_tracking` if `getcert request`
was successful, which is what causes `complete?` to return true.
So any self signed certs would have already been saved off.

lib/manageiq/appliance_console/certificate.rb
jvlcek: Once certs have been successfully retrieved from the authority, `should_request_key?`
will return false consequently avoiding this block and method `save_self_signed` from being invoked.

lib/manageiq/appliance_console/certificate.rb
jvlcek: The `root_filename` file is created by `getcert request` successfully getting signed certs
from the CA and does not exist when only the self signed certs are there.

lib/manageiq/appliance_console/certificate.rb
carbonin: Isn't that not true in the rerun case? When you run `undo_tracking` won't that cause `should_request_key?` to return `true` in which case you will be in this block? Granted, in that case the files will have been deleted by `undo_tracking`, but that brings me back to my original point which is, why save the generated ones off at all?

It seems like you're implementing a lot of logic here which boils down to:
  - delete old certs
  - fetch new certs

Whether they're self signed or not doesn't look like it matters.

jvlcek: @carbonin I `sort of agree`. I agree removing the self signed certs, when not yet
tracking, is fine but I think we should only invoke `undo_tracking` when certmonger
is actively tracking cert, that is when `complete?` is true.

jvlcek:
@carbonin I `think` this is now ready to merge. Thank you for the help.

JoeV

PR #88, opened 2019-05-09
Remove pg cert handling

carbonin:
This removes a case from `appliance_console` that hasn't been accessible since 2014 and removes the cert configuration code from the CLI which hasn't worked since at least 2017 (https://github.com/ManageIQ/manageiq-appliance_console/pull/22).

~~Waiting on approval to remove this, otherwise I will investigate getting it up and running again if we think it's something people want.~~

Approval granted :smile:

### DIFF
--- a/bin/appliance_console
+++ b/bin/appliance_console
@@ -346,27 +346,6 @@ Static Network Configuration
         end
         press_any_key
 
-      when I18n.t("advanced_settings.ca")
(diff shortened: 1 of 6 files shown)


carbonin:
Also this is for https://bugzilla.redhat.com/show_bug.cgi?id=1670150

PR #93, opened 2019-07-05
Fix logfile configuration bz 1720223

bdunne:
DEPENDS ON https://github.com/ManageIQ/linux_admin/pull/208 and a new version of the gem
- unmount before mounting the new device in case there was an old device.
- Update fstab any time we run through this code, the directory may already be listed with a different backing device.
- We failed to match fstab previously because the input was a Pathname and we were comparing to a string
- mount the disk based on the fstab entry

### DIFF
--- a/lib/manageiq/appliance_console/logical_volume_management.rb
+++ b/lib/manageiq/appliance_console/logical_volume_management.rb
@@ -34,8 +34,8 @@ module ApplianceConsole
       create_volume_group
       create_logical_volume_to_fill_volume_group
       format_logical_volume
-      mount_disk
       update_fstab
+      mount_disk
     end
 
     private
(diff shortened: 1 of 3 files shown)


lib/manageiq/appliance_console/logical_volume_management.rb
```diff
@@ -70,6 +70,7 @@ def format_logical_volume
     end
 
     def mount_disk
+      AwesomeSpawn.run!("umount", :params => ["-l", mount_point.to_s]) if File.file?("/proc/mounts") && File.read("/proc/mounts").include?(" #{mount_point} ")
```
carbonin: I think this would be more readable if it were a separate method.

bdunne: Done

bdunne:
@carbonin Everything looks good with the database disk configuration as well

carbonin:
Assuming this is not still WIP?

PR #94, opened 2019-07-08
Add a rubocop channel to the codeclimate config

djberg96:
Add a rubocop channel to the codeclimate config

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -4,15 +4,13 @@ exclude_paths:
 - "**.xml"
 - "**.yaml"
 - "**.yml"
-- "locale/"
-- "spec/"
-- "tools/"
+- locale/
+- spec/
+- tools/
 engines:
   brakeman:
-    # very slow :sad_panda:
     enabled: false
   bundler-audit:
-    # requires Gemfile.lock
     enabled: false
   csslint:
     enabled: false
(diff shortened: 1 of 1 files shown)


carbonin:
Is this supposed to fix the issue with codeclimate or are we missing something still in this repo?

djberg96:
@carbonin Yeah, it was supposed to fix it. Not sure why some of the repos are reaching for rubocop 0.52, though.

carbonin:
> Not sure why some of the repos are reaching for rubocop 0.52, though.

Looks like this PR bundled 0.72.0 so that's not the problem.
https://travis-ci.org/ManageIQ/manageiq-appliance_console/jobs/555747952#L375

carbonin:
Do we need to add `rubocop-performance` somewhere?

djberg96:
@carbonin shouldn't have to, didn't have to for the others. Should be pulled in from the manageiq-content parent, which was updated at https://github.com/ManageIQ/guides/pull/353

djberg96:
@carbonin This is actually a function of the way codeclimate works. If you look at the details of the failure, you'll see it's actually pointing at 0dd3bcf9.

You can that the build for this PR is https://codeclimate.com/github/ManageIQ/manageiq-appliance_console/builds/772, which is green.

I've forwarded you a message from the codeclimate support team that I hope clears things up.

Anyway, you can merge this. :)

PR #96, opened 2019-08-13
Restore the selinux context for the standby data directory

carbonin:
Also fix the data directory permissions.
This makes the standby database disk configuration match the
regular internal database disk configuration more closely

https://bugzilla.redhat.com/show_bug.cgi?id=1740515

### DIFF
--- a/lib/manageiq/appliance_console/database_replication_standby.rb
+++ b/lib/manageiq/appliance_console/database_replication_standby.rb
@@ -62,6 +62,7 @@ module ApplianceConsole
       stop_repmgrd
       initialize_postgresql_disk if disk
       PostgresAdmin.prep_data_directory if disk || resync_data
+      relabel_postgresql_dir
       save_database_yml
       create_config_file(standby_host) &&
         write_pgpass_file &&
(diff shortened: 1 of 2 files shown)


lib/manageiq/appliance_console/database_replication_standby.rb
```diff
@@ -162,8 +163,18 @@ def initialize_postgresql_disk
                                     :volume_group_name   => PostgresAdmin.volume_group_name,
                                     :filesystem_type     => PostgresAdmin.database_disk_filesystem,
                                     :logical_volume_path => PostgresAdmin.logical_volume_path).setup
+
+        # if we mounted the disk onto the postgres user's home directory, fix the permissions
```
jrafanie: > postgres user's home directory

Wouldn't that be /home/postgres ?

carbonin: Nope:
```
[root@localhost ~]# su - postgres -c 'echo $HOME'
/var/lib/pgsql
```

carbonin: Also, this is just a copy of the fix I made in https://github.com/ManageIQ/manageiq-appliance_console/pull/84.

For this bug fix I didn't want to touch the internal database stuff to DRY it up.

carbonin: And also, the real fix is just the `restorecon` part, not sure why we're not having a problem with the directory permissions as well so I figured I would fix this one too.

jrafanie: Could the permissions already be fixed by the above referenced #84 ?  Otherwise, I don't know why it's not complaining.  Postgresql is quite vocal about it's desire specific permissions.

PR #97, opened 2019-09-13
Exit with non-zero when database activation fails

jvlcek:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1747202

Have the cli `exit(1)` in place of `return` when database activation fails. This will  ensure the shell status does not reflects the success code "0"

### To test

Attempt to configure the database using `appliance_console_cli` when it  has already been configured. Then check the shell status for the failed `appliance_console_cli` command with `echo $?` and confirm the shell status is not set to the success code "0"

```bash
/opt/rh/cfme-gemset/bin/appliance_console_cli -i -b /dev/sdb -S -d ${db_name} -U ${db_root} -p '${db_pass}'
echo $?
```

The spec test modifications expect `exit` when database activations fails. This causes `exit` not to actually run in the rspec environment, where it would in actual usage. This results in commands after the `exit` running in rspec environment, which require them to be handled by the test.

### DIFF
--- a/lib/manageiq/appliance_console/cli.rb
+++ b/lib/manageiq/appliance_console/cli.rb
@@ -10,6 +10,8 @@ end
 
 module ManageIQ
 module ApplianceConsole
+  class CliError < StandardError; end
+
   class Cli
     attr_accessor :options
 
@@ -176,6 +178,10 @@ module ApplianceConsole
       install_certs if certs?
       extauth_opts if extauth_opts?
       set_server_state if set_server_state?
+    rescue CliError => e
+      say e
+      say ""
+      exit(1)
     rescue AwesomeSpawn::CommandResultError => e
       say e.result.output
       say e.result.error
@@ -214,17 +220,13 @@ module ApplianceConsole
       # initdb, relabel log directory for selinux, update configs,
       # start pg, create user, create db update the rails configuration,
       # verify, set up the database with region. activate does it all!
-      unless config.activate
-        say "Failed to configure internal database"
-        exit(1)
-      end
+      raise CliError, "Failed to configure internal database" unless config.activate
 
       # enable/start related services
       config.post_activation
     rescue RuntimeError => e
       say e.message
-      say "Failed to configure internal database"
-      exit(1)
+      raise CliError, "Failed to configure internal database"
     end
 
     def set_external_db
@@ -240,10 +242,7 @@ module ApplianceConsole
       }.delete_if { |_n, v| v.nil? })
 
       # call create_or_join_region (depends on region value)
-      unless config.activate
-        say "Failed to configure external database"
-        exit(1)
-      end
+      raise CliError, "Failed to configure external database" unless config.activate
 
       # enable/start related services
       config.post_activation
--- a/lib/manageiq/appliance_console/cli.rb
+++ b/lib/manageiq/appliance_console/cli.rb
@@ -225,8 +225,7 @@ module ApplianceConsole
       # enable/start related services
       config.post_activation
     rescue RuntimeError => e
-      say e.message
-      raise CliError, "Failed to configure internal database"
+      raise CliError, "Failed to configure internal database #{e.message}"
     end
 
     def set_external_db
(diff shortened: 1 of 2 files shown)


jvlcek:
@miq-bot add_label bug
@miq-bot assign @bdunne

jvlcek:
@bdunne Please review

jvlcek:
@kbrock Please review

lib/manageiq/appliance_console/cli.rb
```diff
@@ -216,14 +216,15 @@ def set_internal_db
       # verify, set up the database with region. activate does it all!
       unless config.activate
         say "Failed to configure internal database"
-        return
+        exit(1)
```
kbrock: Why don't you try raising an exception here and rescuing in the run loop ( line ~179)

That way tests can check for an exception raised

jvlcek: Good point @kbrock! Will do.

lib/manageiq/appliance_console/cli.rb
```diff
@@ -176,6 +178,10 @@ def run
       install_certs if certs?
       extauth_opts if extauth_opts?
       set_server_state if set_server_state?
+    rescue CliError => e
+      say(e)
```
kbrock: I was expecting this to be `say e.message`

Do they have the same output?

lib/manageiq/appliance_console/cli.rb
```diff
@@ -10,6 +10,8 @@ def say(arg)
 
 module ManageIQ
 module ApplianceConsole
+  class CliError < StandardError; end
```
kbrock: This works.
Would `ConfigurationError` be easer to read?

lib/manageiq/appliance_console/cli.rb
```diff
@@ -214,16 +220,13 @@ def set_internal_db
...
-      unless config.activate
-        say "Failed to configure internal database"
-        return
-      end
+      raise CliError, "Failed to configure internal database" unless config.activate
 
       # enable/start related services
       config.post_activation
     rescue RuntimeError => e
       say e.message
```
kbrock: I'm pretty sure you can remove this line

and up in the top level rescue add:

```ruby
say e.cause.message if e.cause
```

lib/manageiq/appliance_console/cli.rb
jvlcek: @kbrock Thank you. Yes both `say(e) `and `say(e.message) `produce the same output. I'll update it to `say(e.message)` as I agree it is clearer.

lib/manageiq/appliance_console/cli.rb
jvlcek: I like having the class error exception name match the class name. So I'd like to leave it as `CliError` as it is the error exception for the  `Cli` class. 

Please let me know if you feel strongly about changing the name.


```
...
  class CliError < StandardError; end
      
  class Cli
...
```

lib/manageiq/appliance_console/cli.rb
jvlcek: @kbrock Maybe I don't understand what you are suggesting but I what I `think` you are suggesting is to remove the `raise CliError, "Failed to configure internal database" unless config.activate` that will not work.

If a `RuntimeError` is raised this line will not run but if a `RuntimeError` is not raised having this line is important as it ensure the next line `config.post_activation` does not run.

Please explain.

jvlcek:
@kbrock Thank you for the feedback. Please see my input above.

lib/manageiq/appliance_console/cli.rb
jvlcek: Ah. My bad. I see now that you are suggesting to eliminate one of the invocations to `say()`. Will do. Thank you.

kbrock:
@bdunne you good with this? (I can't merge)

lib/manageiq/appliance_console/cli.rb
kbrock: darn - I could have sworn `cause` existed

PR #104, opened 2019-11-26
Adding Support for OpenID-Connect Authentication configuration from the CLI

abellotti:
Adding support for OpenID-Connect Authentication configuration from the Appliance Console CLI.
    
- Configuring Appliance for OpenID-Connect
``` 
# appliance_console_cli --oidc-config
                        [--oidc-client-host]
                        --oidc-url
                        --oidc-client-id
                        --oidc-client-secret
                        [--oidc-enable-sso]
```
- Unconfiguring Appliance from OpenID-Connect and reverting to Database Authentication
```    
# appliance_console_cli --oidc-unconfig
```


abellotti:
Marking as WIP:

Still needs:
- [ ] testing with OpenID-Connect  (seeing unfamiliar OIDCResponseType error)
- [ ] testing with current SAML CLI config to make sure nothing broke with the refactories
- [ ] adding specs

abellotti:
Related PR:
- https://github.com/ManageIQ/manageiq-appliance/pull/265

abellotti:
@Fryguy when you get a chance and take a 👀 
as discussed moved common code from the SAML bits and used here via AuthUtilities.

/cc @jvlcek

lib/manageiq/appliance_console/auth_utilities.rb
```diff
@@ -0,0 +1,87 @@
+require "uri"
+require "erb"
+
+module ManageIQ
+  module ApplianceConsole
+    module AuthUtilities
+      HTTPD_CONFIG_DIRECTORY = Pathname.new("/etc/httpd/conf.d")
+
+      def restart_httpd
```
jvlcek: This functionality is also done here:
https://github.com/ManageIQ/manageiq-appliance_console/blob/master/lib/manageiq/appliance_console/external_httpd_authentication.rb#L81

That code could be updated to leverage this `AuthUtilities` method. Perhaps that could be done in a separate PR.

lib/manageiq/appliance_console/auth_utilities.rb
```diff
@@ -0,0 +1,87 @@
...
+        @template_directory ||= Pathname.new(ENV.fetch("APPLIANCE_TEMPLATE_DIRECTORY"))
+      end
+
+      def relative_from_root(path)
+        path.absolute? ? path.relative_path_from(Pathname.new("/")) : path
+      end
+
+      # Appliance Settings
+
+      def configure_auth_settings_database
```
jvlcek: Perhaps method `configure_auth_settings_database` could also be used to set openidc and saml, drying up 
https://github.com/ManageIQ/manageiq-appliance_console/blob/master/lib/manageiq/appliance_console/saml_authentication.rb#L169
and
https://github.com/ManageIQ/manageiq-appliance_console/pull/104/files#diff-2778f2fd46589a041acba2350d0b78b1R81

into a single method.

I'd ask you consider renaming `configure_auth_settings_database` to `configure_auth_settings` and
 pass arguments to `configure_auth_settings`, which could dry up `configure_auth_setting_database`,  `configure_auth_setting_openidc` and `configure_auth_settings_saml` into a single method.

e.g.:
```ruby
def configure_auth_settings(args)
  puts("Setting Appliance Authentication Settings to #{args['mode']} ...")
  params = [
    "/authentication/mode=#{args['mode']}",
    "/authentication/httpd_role=#{args['role']}",
    "/authentication/saml_enabled=#{args['saml']}",
    "/authentication/oidc_enabled=#{args['oidc']}",
    "/authentication/sso_enabled=#{args['sso']}",
    "/authentication/provider_type=#{args['type']}"
  ]     

  Utilities.rake_run("evm:settings:set", params)
end   


configure_auth_settings('mode' => 'database',
                        'role' => false,
                        'saml' => false,
                        'oidc' => false,
                        'sso'  => false,
                        'type' => 'none' )

configure_auth_settings('mode' => 'httpd',
                        'role' => true,
                        'saml' => true,
                        'oidc' => false,
                        'sso'  => options[:saml_enable_sso] ? 'true' : 'false',
                        'type' => 'saml' )

configure_auth_settings('mode' => 'httpd',
                        'role' => true,
                        'saml' => false,
                        'oidc' => true,
                        'sso'  => options[:saml_enable_sso] ? 'true' : 'false',
                        'type' => 'oidc' )
```

lib/manageiq/appliance_console/auth_utilities.rb
abellotti: it would be a different say message for the ipa cases,  but yes that too can call AuthUtilities. separate PR.

lib/manageiq/appliance_console/auth_utilities.rb
abellotti: well, I would not get rid of configure_auth_settings_database, because then we're duplicating the logic (what params to set for DB mode) in both SAML and OIDC.  We could keep it, but have it call the generic configure_auth_settings. If we go with a generic configure_auth_settings, we would only build params with what is passed in, just in case another type in the future wants to set a different /authentication/param that's not in saml/oidc/....  The only other thing is losing the friendly message "Setting Appliance Authentication Settings to OpenID-Connect ..."  which is replaced above with httpd, unless we send another param to the generic method for the friendly name, or keep the current ones for the say message, which then call the generic configure_auth_settings for doing the work.

jvlcek: Maybe I'm just missing something but I don’t understand why you don't want to  get rid of configure_auth_settings_database.

My code segment show all three `database`, `oidc` and `saml` can all use 1 generic method. Am I missing something?

I’m not excited about `puts` in this library because the library could be used by the appliance_console and not only the ap_cli. I would ask that you  limit any invocations to `puts` into the cli specific code.

abellotti: what puts ?

jvlcek: Sorry, not puts `say`.  My bad. My point was that I'd rather have a generic utility file not write to stdout. Log to a file perhaps but writing to stdout from a shared utility library doesn't seem like a good idea to me. In this case it could be OK but my philosophy is to avoid writing to stdout from a shared utility.

When setting the authentication configuration, for later use to attempt to diagnose any failures, I'd rather see any message reflect what is actually being set. It's more accurate to log `Setting Appliance Authentication Settings to http` than `Setting Appliance Authentication Settings to OpenID-Connect`. 

Perhaps a:
```ruby
say ("Setting Appliance Authentication Settings for OpenID-Connect support")
configure_auth_settings('mode' => 'httpd',
                        'role' => true,
                        'saml' => false,
                        'oidc' => true,
                        'sso'  => options[:saml_enable_sso] ? 'true' : 'false',
                        'type' => 'oidc' )
```

abellotti: 👍 yep, that would work, but with full parameter names, mode, httpd_role, saml_enabled, etc.  for easier extensibility.

abellotti: @jvlcek 58a21749 pushed as per ^, marked as wip, will do a quick test and unwip if all good.

abellotti: Ok, retested with both OpenID-Connect and SAML, still good. so un-wipping.

Fryguy:
First blush, instead of openidc, should we use the term oidc?  The latter seems more of the industry standard (and more hits on google)

lib/manageiq/appliance_console/auth_utilities.rb
Fryguy: I agree with @jvlcek .  This is duplicated code, and we should DRY it up (I think this PR, but I'm ok with a separate PR).  The message for IPA is barely different, so I don't think that's an issue.

lib/manageiq/appliance_console/auth_utilities.rb
```diff
@@ -0,0 +1,88 @@
...
+        end
+      end
+
+      def path_is_file?(path)
+        path.present? && !path_is_url?(path)
+      end
+
+      def path_is_url?(path)
+        path =~ /\A#{URI.regexp(["http", "https"])}\z/x
+      end
```
Fryguy: These are also duplicate methods FWIW.

lib/manageiq/appliance_console/auth_utilities.rb
```diff
@@ -0,0 +1,88 @@
...
+
+      # File Management
+
+      def remove_file(path)
+        if path.exist?
+          debug_msg("Removing #{path} ...")
+          File.delete(path)
```
Fryguy: Pathname has a delete method directly, so this can be `path.delete`.

lib/manageiq/appliance_console/auth_utilities.rb
```diff
@@ -0,0 +1,88 @@
...
+        end
+      end
+
+      def copy_template(dir, file, template_parameters = nil)
+        src_path = template_directory.join(relative_from_root(dir), file)
+        dest_path = dir.join(file)
+        dest_path = dest_path.sub_ext('') if src_path.extname == ".erb"
+        debug_msg("Copying template #{src_path} to #{dest_path} ...")
+        if src_path.extname == ".erb"
```
Fryguy: I don't understand why there's .erb templating here, but not with other things like saml.

lib/manageiq/appliance_console/auth_utilities.rb
```diff
@@ -0,0 +1,88 @@
...
+        configure_auth_settings(:mode          => "database",
+                                :httpd_role    => false,
+                                :saml_enabled  => false,
+                                :oidc_enabled  => false,
+                                :sso_enabled   => false,
+                                :provider_type => "none")
+      end
+
+      def configure_auth_settings(args)
+        Utilities.rake_run("evm:settings:set", args.collect { |key, val| "/authentication/#{key}=#{val}" })
```
Fryguy: do we have to be careful with command line injection issues here?  I didn't look at what rake_run does.

Fryguy:
Looking really good!

lib/manageiq/appliance_console/auth_utilities.rb
abellotti: SAML auth handshaking (certs/keys) is file driven, OIDC id/secrets are in the conf file so erb to the rescue.

lib/manageiq/appliance_console/auth_utilities.rb
```diff
@@ -0,0 +1,88 @@
...
+        src_path = template_directory.join(relative_from_root(dir), file)
+        dest_path = dir.join(file)
+        dest_path = dest_path.sub_ext('') if src_path.extname == ".erb"
+        debug_msg("Copying template #{src_path} to #{dest_path} ...")
+        if src_path.extname == ".erb"
+          template = ERB.new(File.read(src_path), nil, '-')
+          if template_parameters
+            File.write(dest_path, template.result_with_hash(template_parameters))
+          else
+            File.write(dest_path, template.result(binding))
```
Fryguy: I would get rid of this part...maybe raise an ArgumentError exception instead if .erb is given but template_parameters is not passed.

abellotti:
@Fryguy  updates done  --openidc-* => --oidc-* parameters updated with related changes to the classes

jvlcek:
@abellotti and @Fryguy 

Just an FYI, I've tested these changes on a live MiQ appliance to successfully configure  OIDC.

abellotti:
Thanks @jvlcek for kicking the tire 👍

lib/manageiq/appliance_console/auth_utilities.rb
Fryguy: Discussed with @abellotti and this looks good.

PR #106, opened 2019-12-19
update hostname regex to allow starting with digit

d-m-u:
Hostnames can start with a digit per https://tools.ietf.org/html/rfc1123, and has a restriction on length that we're not handling, and should be updated. 

@yrudman please review? i and regexes are not friends

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1785257

### DIFF
--- a/lib/manageiq/appliance_console/prompts.rb
+++ b/lib/manageiq/appliance_console/prompts.rb
@@ -13,7 +13,7 @@ module ApplianceConsole
     DOMAIN_REGEXP = /^[a-z0-9]+([\-\.]{1}[a-z0-9]+)*(\.[a-z]{2,13})?$/
     INT_REGEXP    = /^[0-9]+$/
     NONE_REGEXP   = /^('?NONE'?)?$/i.freeze
-    HOSTNAME_REGEXP = /^(([a-zA-Z]|[a-zA-Z][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z]|[A-Za-z][A-Za-z0-9\-]*[A-Za-z0-9])$/
+    HOSTNAME_REGEXP = /^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$/.freeze
 
     def ask_for_uri(prompt, expected_scheme, opts = {})
       require 'uri'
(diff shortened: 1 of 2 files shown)


d-m-u:
@miq-bot add_label hammer/yes, ivanchuk/yes
@miq-bot add_label bug

yrudman:
@d-m-u I think whole expression could be simplified.
You can use some tool to test and get explanation: https://regexr.com/

lib/manageiq/appliance_console/prompts.rb
```diff
@@ -13,7 +13,7 @@ module Prompts
     DOMAIN_REGEXP = /^[a-z0-9]+([\-\.]{1}[a-z0-9]+)*(\.[a-z]{2,13})?$/
     INT_REGEXP    = /^[0-9]+$/
     NONE_REGEXP   = /^('?NONE'?)?$/i.freeze
-    HOSTNAME_REGEXP = /^(([a-zA-Z]|[a-zA-Z][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z]|[A-Za-z][A-Za-z0-9\-]*[A-Za-z0-9])$/
+    HOSTNAME_REGEXP = /^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z]|[A-Za-z][A-Za-z0-9\-]*[A-Za-z0-9])$/
```
carbonin: I think the second part needs to be updated too.
Grabbed this from https://stackoverflow.com/questions/106179/regular-expression-to-match-dns-hostname-or-ip-address

`^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$`

carbonin:
Also can you add a spec for the new behavior?

spec/prompts_spec.rb
```diff
@@ -153,9 +153,15 @@
         expect(subject.ask_for_ip('prompt', '1.1.1.1')).to eq('dead:beef::1')
       end
 
-      it "should handle hostname" do
-        say %w(5x redhat.com)
-        expect(subject.ask_for_ip_or_hostname("prompt", "1.1.1.1")).to eq("redhat.com")
+      it "should handle hostname beginning with a digit" do
+        say "5x"
```
carbonin: I would prefer we change this spec to something that actually looks more like an actual hostname. How about `198.51.100.1.example.com`

d-m-u:
@simaishi the bug opener wants this backported to 5.10 and 5.11, can we do that please?

simaishi:
@d-m-u That's not up to me to decide... but you can start by putting the BZ to POST 😄

carbonin:
This was actually not included in a release of the gem yet so the bug will still be present in `master`.

carbonin:
I can release a new version today though.

PR #107, opened 2019-12-28
Use the regex to validate hostnames when we set them

d-m-u:
We only validate hostnames today when you test them (options 1 and 4 in the console), not when you're actually applying them (options 1 and 5), so currently you can verify that ```_test1``` shouldn't be valid due to the underscore but you will be able to set it anyway. 

[Highline's](https://github.com/JEG2/highline/blob/bf489b632afd92c68df0b2c1f348afb66b5495aa/lib/highline/question.rb#L144) [third arg for this method](https://github.com/ManageIQ/manageiq-appliance_console/blob/7372c5b864fc76494708362e4a68dc1f00234680/lib/manageiq/appliance_console/prompts.rb#L192) is [the validation](https://github.com/ManageIQ/manageiq-appliance_console/pull/106/files#diff-11f3fdac8dc445b3e93dd2459f62b9a9L16) so we should use it. 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1786282

@kbrock sorry, another super trivial one, but any chance I could get you to review please? 

Behavior is all tested already, see [the specs on 106](https://github.com/ManageIQ/manageiq-appliance_console/pull/106)

## edit:
Screen Shot 2019-12-28 at 2 32 13 PM
#### welp okay cool then 
I mean that's about how I feel too so

### DIFF
--- a/bin/appliance_console
+++ b/bin/appliance_console
@@ -280,7 +280,7 @@ Static Network Configuration
 
         when 'hostname'
           say("Hostname Configuration\n\n")
-          new_host = just_ask("new hostname", host)
+          new_host = just_ask("new hostname", host, HOSTNAME_REGEXP, "Please enter a valid hostname")
 
           if new_host != host
             say("Applying new hostname...")


d-m-u:
@miq-bot add_label bug

bin/appliance_console
```diff
@@ -280,7 +280,7 @@ Static Network Configuration
 
         when 'hostname'
           say("Hostname Configuration\n\n")
-          new_host = just_ask("new hostname", host)
+          new_host = just_ask("new hostname", host, HOSTNAME_REGEXP)
```
kbrock: Is there an error message that would be more helpful than "In the specified format"?

d-m-u: Sure, whatcha think about ```"Please provide a text string up to 24 characters drawn from the alphabet (A-Z), digits (0-9), minus sign (-), and period (.) per IETF requirements."``` ?

d-m-u: Also, since I'm gone for two weeks, any chance you'd be amenable to merging this and letting me do the error text update in a follow up when I get back, please?

kbrock: um yea, think it is fine the way it is.

bdunne: > Sure, whatcha think about `"Please provide a text string up to 24 characters drawn from the alphabet (A-Z), digits (0-9), minus sign (-), and period (.) per IETF requirements."` ?

Maybe "Please enter a valid hostname"?

kbrock: I like Brandon's suggestion.

Either that or:

>Please pick a valid host name according to the IETF standard and not something esoteric like having a space or other non alphanumeric character. Sure, minus signs are acceptable and if you want to fully qualify it then by all means add a period. But if you are going to do that then please put in a proper TLD (e.g. .com) and don't put in some bogus string that will puke when we send it over to dns. Thanks

PR #108, opened 2020-01-14
Deduplicate the SCAP_RULES_DIR constant

carbonin:
642781d50 added a second definition of the constant when adding
the scap feature to the cli. This caused a warning on console startup:

manageiq-appliance_console-5.3.0/bin/appliance_console:22: warning: already initialized constant SCAP_RULES_DIR
manageiq-appliance_console-5.3.0/lib/manageiq/appliance_console/cli.rb:12: warning: previous definition of SCAP_RULES_DIR was here

Instead of re-defining the constant we can move it into the Scap
class and make it the default parameter value for #initialize

### DIFF
--- a/bin/appliance_console
+++ b/bin/appliance_console
@@ -19,8 +19,6 @@ include HighLine::SystemExtensions
 
 require 'manageiq/appliance_console/i18n'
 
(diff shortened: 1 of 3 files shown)


yrudman:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1793070

PR #111, opened 2020-03-03
Use LinuxAdmin::Service#start with enable set to true

carbonin:
This will enable and start the service in one command, hopefully
avoiding whatever is causing https://github.com/ManageIQ/manageiq/issues/10300


carbonin:
https://github.com/ManageIQ/manageiq-appliance_console/pull/112 should fix the spec failures.

PR #113, opened 2020-03-24
Add net-ssh gems needed to support ed25519 cipher

jvlcek:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1815568

net-ssh requires the following gems for SSH ed25519 cipher support:
- pbnacl (>= 3.2, < 5.0)
- pbnacl-libsodium
- berypt_pbkdf (>= 1.8, < 2.8)

This PR adds the needed gems.

**To test:**

_Gen the ed25519 key:_
> miq-host> ssh-keygen -o -a 256 -t ed25519 -C "$(hostname)-$(date +'%d-%m-%Y')"

This produces:
  ~/.ssh/id_ed25519
  ~/.ssh/id_ed25519.pub

To use this put the content of the ~/.ssh/id_ed25519.pub in the ~/.ssh/authorized_keys on the remote host

_Then ensure the following command runs successfully:_

> miq-host> appliance_console_cli --fetch-key=_remote hostname or IPAddr_


jvlcek:
@miq-bot add_label bug
@miq-bot assign @bdunne

jvlcek:
@bdunne  and @simaishi Please take a look.

I tested on a live appliance by using an ed25519 cipher. I saw the appliance console cli command fail as described in the BZ. Then applied the needed gems as described here 
https://github.com/net-ssh/net-ssh/issues/478 and was able to successfully run the previously failing command.

jrafanie:
@jvlcek are any of these compiled gems requiring system libraries?  I know bcrypt itself is but not sure about bcrypt_pbkdf or the others.  If they require system libraries, we'll have to make sure @simaishi can package them and get the correct version dependencies.

Fryguy:
@jvlcek This PR looks good, but is it possible to add specs for something like this?  I only ask because if someone tries to remove these gems in the future, a spec would prevent them (and would effectively document why they are needed).

jrafanie:
Note, is this an enhancement or a bug? Are we adding additional support or did something break that requires these new gems?  It seems odd to add gems to "fix" a bug.

Maybe it's a bughancement.

simaishi:
@jvlcek so we don't need to add net-scp/net-ssh?

We discussed a bit in gitter, but I'm not sure exactly what gems are supposed to be in .gemspec here. A code in this repo uses Net::SCP, which uses Net::SSH, which needs those 3 extra gems. 

The code works now on appliance because manageiq-gems-pending brings in net-scp/net-ssh anyway, but I thought we need to add at least net-scp since manageiq-applaince_console directly depends on it.  It looks strange to add just 3 gems. But I'm not sure what we do for cases like this...

jvlcek:
@jrafanie Yes bcrypt_pdkfd is compiled. @simaishi Can you please comment on what would be required to ensure they can be properly compiled?

jvlcek:
@simaishi Good point but there could be other gems that manageiq-gems-pending brings in that this code needs. I'm not sure if this is the place to address that. This PR is focused on providing the ability to enable net-ssh to read password encrypted Ed25519 keys.

If you disagree please let me know. Thank you! JoeV

simaishi:
> @simaishi Good point but there could be other gems that manageiq-gems-pending brings in that this code needs. I'm not sure if this is the place to address that. This PR is focused on providing the ability to enable net-ssh to read password encrypted Ed25519 keys.

I mentioned because net-scp/net-ssh is directly related to this PR. But like I said, I'm not even sure what should be in .gemspec, so I'll let others discuss that.

> @jrafanie Yes bcrypt_pdkfd is compiled. @simaishi Can you please comment on what would be required to ensure they can be properly compiled?

rbnacl-ibsodium is also a compiled gem. Since this change is needed in ivanchuk release, I'll need to start process of crating RPMs for those 2 gems for downstream.

jvlcek:
@Fryguy I'm not sure how to add a spec for this. These additional gems are used by net-ssh to read Ed25519 encrypted keys. I guess I can explore having a spec generate an Ed25519 ssh key and attempt to use it. I may reach out for help figuring how and if that would be worth doing.

jvlcek:
@simaishi, @jrafanie and @Fryguy Thank you for the help/input on this!

jvlcek:
> > @simaishi Good point but there could be other gems that manageiq-gems-pending brings in that this code needs. I'm not sure if this is the place to address that. This PR is focused on providing the ability to enable net-ssh to read password encrypted Ed25519 keys.
> 
> I mentioned because net-scp/net-ssh is directly related to this PR. But like I said, I'm not even sure what should be in .gemspec, so I'll let others discuss that.

Correct but this PR is not adding net-scp/net-ssh usage. That usage has already existed. This PR just provides net-scp support for Ed25519 keys.

bdunne:
All of our direct dependencies should be listed in the gemspec.  Please add `net-scp` to the gemspec as a runtime dependency

jrafanie:
> Correct but this PR is not adding net-scp/net-ssh usage. That usage has already existed. This PR just provides net-scp support for Ed25519 keys.

Right.  If console was implicitly depending on another repo to provide net-ssh gem, we should fix that.  Since this PR is directly related to extending net-ssh support, it might make sense to make this dependency on net-ssh explicit by copying the same net-ssh requirement (as we have elsewhere) to this gemspec.

Note, if this PR did not have much to do with net-ssh, I would agree, you could add the net-ssh dependency outside of the PR, but since this PR is about net-ssh and you found that console is not stating it's dependency on net-ssh properly, it makes sense to do it here.

jvlcek:
@bdunne and @jrafanie Good points! Thank you will do.

jrafanie:
I'd still like confirmation that we have the required devel packages to build these gems on rhel/centos but looks good to me other than that.

jvlcek:
> I'd still like confirmation that we have the required devel packages to build these gems on rhel/centos but looks good to me other than that.

@jrafanie 
Thank you for helping me check this offline. Seems we do have the required devel packages to build these gems.

Do you agree?

jrafanie:
I agree, I'd like @simaishi's final 👍

simaishi:
Trying to create RPM for `rbnacl-libsodium` for downstream and noticed this at:
https://github.com/RubyCrypto/rbnacl-libsodium

```
THIS GEM HAS OFFICIALLY BEEN DISCONTINUED.

Please use your operating system's package manager to install libsodium when using RbNaCl. For more information, see:

https://github.com/crypto-rb/rbnacl/wiki/Installing-libsodium
```

jvlcek:
> Should this be an optional thing where users will need to install `libsodium` if they want ed25519 cipher support? Upstream appliance has it already installed.

Hey @simaishi,

Except for net-scp I don't see any of the gems this PR added installed on upstream manageiq prior to this PR.

> miq> bundle info libsodium
> Could not find gem 'libsodium'.
>
> miq> bundle info rbnacl-libsodium
> Could not find gem 'rbnacl-libsodium'.
>
> miq> bundle info rbnacl
> Could not find gem 'rbnacl'.
>
> miq> bundle info bcrypt_pbkdf
> Could not find gem 'bcrypt_pbkdf'.
> 

Maybe I'm missing something. Please explain.

@bdunne, @Fryguy and @jrafanie 

Do we deliver anything else that requires customers to manually install gems needed for specific sub-functionality?

simaishi:
> Except for net-scp I don't see any of the gems this PR added installed on upstream manageiq prior to this PR.

Not `libsodium` gem, but the library which can come from RPM.

> Do we deliver anything else that requires customers to manually install gems needed for specific sub-functionality?

In this case, "customers" is really just developers who run code without using appliance, as appliance already has libsodium rpm installed.

simaishi:
I tried without `rbnacl-libsodium` gem on an upstream appliance, but it failed to find `libsodium.so`. The libsodium rpm puts `libsodium.so.23` and `libsodium.so.23.3.0` on the machine, but not `libsodium.so` and ffi fails to find the library.

`libsodium-devel` rpm puts libsodium.so and it will work, but that will be a problem for downstream.  I went ahead and created rbnacl-libsodium gem RPM for downstream.

PR #114, opened 2020-04-07
Messaging configuration

bdunne:
Initial effort for configuring an external queue / messaging system.
TODO:
- [x] add tests
- [x] test on an appliance


bin/appliance_console
```diff
@@ -411,6 +412,18 @@ Static Network Configuration
         # Get the region again because it may have changed
         region = ManageIQ::ApplianceConsole::DatabaseConfiguration.region
 
+        queue_options = {
+          "Use an external queue" => "external_queue",
```
Fryguy, line 416: Let's use the term "messaging system", I'm thinking, even calling this external_messaging

lib/manageiq/appliance_console/cli.rb
```diff
@@ -426,6 +435,16 @@ def set_server_state
     def self.parse(args)
       new.parse(args).run
     end
+
+    def set_external_db
```
Fryguy, line 439: Typo? set_external_messaging?

lib/manageiq/appliance_console/i18n.rb
```diff
@@ -6,7 +6,7 @@
 locales_dir = ENV['CONTAINER'] ? "container" : "appliance"
 locales_paths = [
   gem_root.join("locales", locales_dir, "*.yml"),
-  File.expand_path(File.join("productization/appliance_console/locales", locales_dir, "*.yml"), ManageIQ::ApplianceConsole::RAILS_ROOT)
+  # File.expand_path(File.join("productization/appliance_console/locales", locales_dir, "*.yml"), ManageIQ::ApplianceConsole::RAILS_ROOT)
```
Fryguy, line 9: ?

bin/appliance_console
chessbyte: Really good explanation of message queues from Microsoft [here](https://docs.microsoft.com/en-us/previous-versions/msp-n-p/dn589781(v=pandp.10)).  Agree that the critical term is messaging, not queueing.

lib/manageiq/appliance_console/i18n.rb
bdunne: That was for testing, will be removed.

bin/appliance_console
bdunne: :+1:

bin/appliance_console
```diff
@@ -383,7 +383,8 @@ Static Network Configuration
           "Create Internal Database"           => "create_internal",
           "Create Region in External Database" => "create_external",
           "Join Region in External Database"   => "join_external",
-          "Reset Configured Database"          => "reset_region"
+          "Reset Configured Database"          => "reset_region",
+          "No changes"                         => "no_changes"
```
carbonin, line 387: Why add this? Is this just in case someone wants to configure kafka but not the database? Would it be less confusing as a new top-level item or do we want to keep all the "initial setup"-type stuff in one wizard-style menu?

carbonin: Also, if we're keeping it this way, typically we ask the user for all of the information then offer a summary. Only then do we actually go and do any of the work.

Do we want to ask for the database info, then ask for the kafka info, then do whatever we need to?

bdunne: I think I discussed the flow with @Fryguy a while ago and we thought that the database and queue configuration should be tightly coupled and that this should be "application configuration" as opposed to the other things in the menu which are OS configuration.  So, I thought it would be a good idea to be able to skip database changes if you only wanted to make queue changes.

bdunne: > Do we want to ask for the database info, then ask for the kafka info, then do whatever we need to?

Probably a good idea.  I'll change it

bdunne:
@carbonin @Fryguy @chessbyte Any other concerns?

carbonin:
The only thing I would mention is that right now the database configuration raises when `run_interactive` fails. That would mean that we wouldn't get to the kafka config even if it could have succeeded. This might be fine for now, but I just wanted to bring it up.

bdunne:
> The only thing I would mention is that right now the database configuration raises when `run_interactive` fails. That would mean that we wouldn't get to the kafka config even if it could have succeeded. This might be fine for now, but I just wanted to bring it up.

I think this is okay.  What do you think?

carbonin:
*shrug* I'm fine with it.

Fryguy:
Just getting back to this, but looks good to me.  @bdunne I think we need to update downstream since you changed the I18n key.  Additionally, this might require changes to documentation.

PR #115, opened 2020-04-08
Top level constant warnings in specs

bdunne:
Remove top-level constant causing warnings

```
spec/database_replication_spec.rb:2: warning: already initialized constant SPEC_NAME
spec/database_replication_primary_spec.rb:2: warning: previous definition of SPEC_NAME was here
spec/database_replication_standby_spec.rb:2: warning: already initialized constant SPEC_NAME
spec/database_replication_spec.rb:2: warning: previous definition of SPEC_NAME was here
```


carbonin:
@bdunne I think you can probably just remove all the spec_name stuff, right? I'm not sure why it's needed...

PR #117, opened 2020-05-08
Add support for the oidc introspection endpoint

jvlcek:
Fixes ManageIQ/manageiq#19866

This PR adds support for the oidc introspection endpoint to the appliance console CLI. Combined with other PRs in other github repos support for the ManageIQ API  will be moved out of code and into the Apache configuration.

**Dependent PRs**
https://github.com/ManageIQ/manageiq-appliance/pull/282
https://github.com/ManageIQ/manageiq-api/pull/828
https://github.com/ManageIQ/manageiq/pull/20131

To Test:

Run the command of the form

```bash
appliance_console_cli \
  --oidc-config \
  --oidc-url="http://my-keycloak.example.com:8080/auth/realms/miq/.well-known/openid-configuration" \
  --oidc-client-host="my-miq.example.com" \
  --oidc-client-id="my-oidc-client"  \
  --oidc-client-secret="********" \
  --oidc-introspection-endpoint="https://my-keycloak.example.com:8443/auth/realms/miq/protocol/openid-connect/token/introspect"
```

### DIFF
--- a/lib/manageiq/appliance_console/oidc_authentication.rb
+++ b/lib/manageiq/appliance_console/oidc_authentication.rb
@@ -52,10 +52,12 @@ module ManageIQ
         debug_msg("Copying Apache OpenID-Connect Config files ...")
         copy_template(HTTPD_CONFIG_DIRECTORY, "manageiq-remote-user-openidc.conf")
         copy_template(HTTPD_CONFIG_DIRECTORY, "manageiq-external-auth-openidc.conf.erb",
-                      :miq_appliance              => host,
-                      :oidc_provider_metadata_url => options[:oidc_url],
-                      :oidc_client_id             => options[:oidc_client_id],
-                      :oidc_client_secret         => options[:oidc_client_secret])
+                      :miq_appliance               => host,
+                      :oidc_provider_metadata_url  => options[:oidc_url],
+                      :oidc_client_id              => options[:oidc_client_id],
+                      :oidc_client_secret          => options[:oidc_client_secret],
+                      :oidc_introspection_endpoint => options[:oidc_introspection_endpoint],
+)
       end
 
       def remove_apache_oidc_configfiles
@@ -74,6 +76,7 @@ module ManageIQ
         raise "Must specify the OpenID-Connect Provider URL via --oidc-url" if options[:oidc_url].blank?
         raise "Must specify the OpenID-Connect Client ID via --oidc-client-id" if options[:oidc_client_id].blank?
         raise "Must specify the OpenID-Connect Client Secret via --oidc-client-secret" if options[:oidc_client_secret].blank?
+        raise "Must specify the OpenID-Connect Client Introspection Endpoint --oidc-introspection-endpoint" if options[:oidc_introspection_endpoint].blank?
       end
 
       # Appliance Settings
--- a/lib/manageiq/appliance_console/oidc_authentication.rb
+++ b/lib/manageiq/appliance_console/oidc_authentication.rb
@@ -5,9 +5,9 @@ module ManageIQ
 
       attr_accessor :host, :options
 
-      URL_SUFFIX = /\/\.well-known\/openid-configuration$/
-      INTROSPECT_SUFFIX = "/protocol/openid-connect/token/introspect"
-      INTROSPECT_ENDPOINT_ERROR = "Unable to derive the OpenID-Connect Client Introspection Endpoint. Use --oidc-introspection-endpoint"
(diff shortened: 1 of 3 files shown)


Fryguy:
All dependent PRs merged.

lib/manageiq/appliance_console/oidc_authentication.rb
```diff
@@ -74,6 +76,7 @@ def validate_oidc_options
         raise "Must specify the OpenID-Connect Provider URL via --oidc-url" if options[:oidc_url].blank?
         raise "Must specify the OpenID-Connect Client ID via --oidc-client-id" if options[:oidc_client_id].blank?
         raise "Must specify the OpenID-Connect Client Secret via --oidc-client-secret" if options[:oidc_client_secret].blank?
+        raise "Must specify the OpenID-Connect Client Introspection Endpoint --oidc-introspection-endpoint" if options[:oidc_introspection_endpoint].blank?
```
Fryguy, line 79: Why are we making it required?  I would expect that most providers have a well known endpoint, so I would argue this is optional, and if specified then we include the line.  @abellotti Thoughts?

jvlcek:
@abellotti and @Fryguy  oidc_introspection_endpoint is no longer required if the metadata URL can be used to derive it. I think I have addressed all the issues that have been raised.

Thank you again for the input.

lib/manageiq/appliance_console/oidc_authentication.rb
abellotti: Any objections/issues to having the CLI attempt to fetch the provider metadata at the url specified ?  If we can, then we can simply get the exact introspection URL needed, this would work for all providers and not just keycloak (as well as future IAM where the provider metadata would be specified).

so enhanced logic would be:

- if introspection url not specified:
  - try to fetch the provider metadata, if successful, use the introspection value there.
  - if unsuccessful, try to derive as done in this PR.
  - if cannot derive, fail by requiring the option.

jvlcek: > Any objections/issues to having the CLI attempt to fetch the provider metadata at the url specified ? If we can, then we can simply get the exact introspection URL needed, this would work for all providers and not just keycloak (as well as future IAM where the provider metadata would be specified).
> 
> so enhanced logic would be:
> 
> * if introspection url not specified:
>   
>   * try to fetch the provider metadata, if successful, use the introspection value there.
>   * if unsuccessful, try to derive as done in this PR.
>   * if cannot derive, fail by requiring the option.

Are you suggesting something like this:

```bash
curl -k -L -X GET     -H 'Content-Type: application/x-www-form-urlencoded'     -d 'grant_type=password'     -d 'client_id=${options[:oidc_client_id]}'     -d 'client_secret=${options[:oidc_client_secret]}'  ${options[:oidc_url]}  | sed -e 's/[{}]/''/g'  | sed -s 's/.*introspection_endpoint\":\"//' | sed -s 's/.\".*$//'
```

I do not think we should rely on something like `jq` being there to parse the JSON.

I'm a little concerned that parsing the JSON might be more trouble than the outcome would be worth.

Thoughts?

Fryguy:
@jvlcek @abellotti If a user does _not_ pass the token introspection endpoint and we can derive it, do we still put that value in the conf file?  Is it required in the conf file?

abellotti:
I think so, The spec says it is mandatory here https://github.com/zmartzone/mod_auth_openidc/blob/9f1881fb07b8c4fbf8d5a4ce9be6c191f812befb/auth_openidc.conf#L320

@jvlcek have we tried without it  but with the OIDCOAuthClientID and OIDCOAUTHClientSecret specified ?

jvlcek:
> I think so, The spec says it is mandatory here https://github.com/zmartzone/mod_auth_openidc/blob/9f1881fb07b8c4fbf8d5a4ce9be6c191f812befb/auth_openidc.conf#L320
> 
> @jvlcek have we tried without it but with the OIDCOAuthClientID and OIDCOAUTHClientSecret specified ?

I tried it just now, just to be sure.

Yes OIDCOAuthIntrospectionEndpoint is mandatory. 
Things fail miserably without   OIDCOAuthIntrospectionEndpoint even if OIDCOAuthClientID and OIDCOAUTHClientSecret are specified.

Fryguy:
@bdunne Can you cut a release of this to include this change?

bdunne:
> @bdunne Can you cut a release of this to include this change?

Released v5.5.0

chessbyte:
@Fryguy @bdunne @jvlcek @abellotti Created https://github.com/ManageIQ/manageiq-appliance/pull/284 to use the newly released gemfile.

Issue #119, opened 2020-05-12
Fetch the introspection endpoint from the provider metadata


abellotti:
For the new :oidc_introspection_endpoint option, if not specified we should attempt to fetch its value from the provider metadata obtained from the metadata url specified.  This would guarantee the exact value needed for Keycloak and all other providers so we no longer need to derive the value.

if introspection url is not specified:
- try to fetch the provider metadata, if successful, use the introspection value defined there.
- If we can't fetch the provider metadata, fail by requiring the :oidc_introspection_endpoint option.

abellotti:
follow-up to https://github.com/ManageIQ/manageiq-appliance_console/pull/117

jvlcek:
@miq-bot assign @jvlcek

PR #121, opened 2020-05-13
Try to fetch introspect endpoint from the provider metadata

jvlcek:
Fixes https://github.com/ManageIQ/manageiq-appliance_console/issues/119

This PR adds support to attempt to fetch the introspection endpoint by querying the provider metadata URL.

The logic for deriving the introspection URL used to be:

```
if introspection url is not specified:

try to derive the introspection value from the the metadata url value
If not possible, fail by requiring the :oidc_introspection_endpoint option.
```
The new logic for deriving the introspection URL will become:

```
if introspection url is not specified:

try to fetch the introspection endpoint from the provider metadata
if not successful fail by requiring the :oidc_introspection_endpoint option.
```

### DIFF
--- a/lib/manageiq/appliance_console/oidc_authentication.rb
+++ b/lib/manageiq/appliance_console/oidc_authentication.rb
@@ -66,10 +66,12 @@ module ManageIQ
                       :oidc_client_secret          => options[:oidc_client_secret],
                       :oidc_introspection_endpoint => options[:oidc_introspection_endpoint])
 
-        File.open("#{HTTPD_CONFIG_DIRECTORY}/manageiq-external-auth-openidc.conf", "a") do |f|
-          f.write("\nOIDCSSLValidateServer      Off\n")
-          f.write("OIDCOAuthSSLValidateServer Off\n")
-        end if options[:oidc_insecure]
+        if options[:oidc_insecure]
+          File.open("#{HTTPD_CONFIG_DIRECTORY}/manageiq-external-auth-openidc.conf", "a") do |f|
+            f.write("\nOIDCSSLValidateServer      Off\n")
+            f.write("OIDCOAuthSSLValidateServer Off\n")
+          end
+        end
       end
 
       def remove_apache_oidc_configfiles
--- a/lib/manageiq/appliance_console/oidc_authentication.rb
+++ b/lib/manageiq/appliance_console/oidc_authentication.rb
@@ -65,6 +65,11 @@ module ManageIQ
                       :oidc_client_id              => options[:oidc_client_id],
                       :oidc_client_secret          => options[:oidc_client_secret],
                       :oidc_introspection_endpoint => options[:oidc_introspection_endpoint])
+
+        File.open("#{HTTPD_CONFIG_DIRECTORY}/manageiq-external-auth-openidc.conf", "a") do |f|
+          f.write("\nOIDCSSLValidateServer      Off\n")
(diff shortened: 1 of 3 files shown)


jvlcek:
@miq-bot assign @abellotti

lib/manageiq/appliance_console/oidc_authentication.rb
```diff
@@ -85,10 +88,26 @@ def validate_oidc_options
...
+        http.use_ssl = (uri.scheme == "https")
+
+        request = Net::HTTP::Get.new(uri.request_uri)
+        request.basic_auth(options[:oidc_client_id], options[:oidc_client_secret])
+        response = http.request(request)
+
+        JSON.parse(response.body)["introspection_endpoint"]
+      end
+
+      def compose_introspection_endpoint
```
abellotti, line 107: let's drop the derivation logic, we with can fetch it or they specify it.

lib/manageiq/appliance_console/oidc_authentication.rb
```diff
@@ -85,10 +88,26 @@ def validate_oidc_options
...
 
-        options[:oidc_introspection_endpoint] = options[:oidc_url].sub(URL_SUFFIX, INTROSPECT_SUFFIX) if options[:oidc_url].match(URL_SUFFIX)
+        options[:oidc_introspection_endpoint] = fetch_introspection_endpoint || compose_introspection_endpoint
         raise INTROSPECT_ENDPOINT_ERROR if options[:oidc_introspection_endpoint].blank?
       end
 
+      def fetch_introspection_endpoint
+        uri = URI.parse(options[:oidc_url])
+        http = Net::HTTP.new(uri.host, uri.port)
+        http.use_ssl = (uri.scheme == "https")
```
abellotti, line 98: Will this work with self signed cert in local keycloak setup ? we may need to add http.verify_mode = OpenSSL::SSL::VERIFY_NONE with an option like --insecure

lib/manageiq/appliance_console/oidc_authentication.rb
```diff
@@ -85,10 +88,26 @@ def validate_oidc_options
...
+
+        request = Net::HTTP::Get.new(uri.request_uri)
+        request.basic_auth(options[:oidc_client_id], options[:oidc_client_secret])
+        response = http.request(request)
+
+        JSON.parse(response.body)["introspection_endpoint"]
```
abellotti, line 104: probably need a rescue for above logic and return nil on failure.

lib/manageiq/appliance_console/oidc_authentication.rb
jvlcek: Correct. this will not work with self signed cert and I am concerned  it would be safe to use `http.verify_mode = OpenSSL::SSL::VERIFY_NONE`.
Which is why I didn't drop the `compose_introspection_endpoint`.

@abellotti Do you think using  OpenSSL::SSL::VERIFY_NONE is safe?

abellotti: Of course not, default must always be on.    Can't we make it optional for dev setups (i.e. via an option like --insecure or something similar to that) ?

abellotti:
Thanks @jvlcek for the enhancement, LGTM!! 👍

For the allow_any_instance_of warnings, we can probably use web mock to instrument the http gets, but that can be done as a follow-up.

Issue #125, opened 2020-10-13
Automate Kafka configuration in the Appliance Console


jvlcek:
This issue outlines the manual steps required to configure Kafka on a server and client.

This issue will be used to track automating these steps in the Appliance console.

#  **On the Kafka server VM:**

### PRE-STEPs Server Setup: Set the hostname and install KAFKA RPM for now. Won't need this later.

```bash
      my-kafka-server> hostnamectl set-hostname my-kafka-server.example.redhat.com
      my-kafka-server> yum install kafka # for now need to manually install kafka
```

### STEP 1: Get the hostname

```bash
       hostname -f
```

### STEP 2: Gather User supplied data

      _username, password..._

### STEP 3: Create the log4 configuration file

```bash
      my-kafka-server> cp /opt/kafka/config-sample/tools-log4j.properties /opt/kafka/config/tools-log4j.properties 
```

### STEP 4: Create the kafka server jaas config file

```bash
      my-kafka-server 192.0.2.78> cat /opt/kafka/config/kafka_server_jaas.conf 
      KafkaServer {
         org.apache.kafka.common.security.plain.PlainLoginModule required
         username="my-user"
         password="********"
         user_admin="********";
      };
```

### STEP 5: Update /usr/lib/systemd/system/kafka.service to include the **KAFKA_OPTS NOT NEEDED AFTER MY  PR SHOWS UP IN THE BUILD**  

```bash
        [Service]
      + Environment="KAFKA_OPTS=-Djava.security.auth.login.config=/opt/kafka/config/kafka_server_jaas.conf"
        ExecStart=/opt/kafka/bin/kafka-server-start.sh /opt/kafka/config/server.properties
```

### STEP 6: Create the client.properties config 

```bash
      my-kafka-server 192.0.2.78> cat client.properties
      ssl.truststore.location=/opt/kafka/config/abc/truststore.jks
      ssl.truststore.password=********

      sasl.mechanism=PLAIN
      security.protocol=SASL_SSL
      sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required \
        username="my-user" \
        password="********";
```

### STEP 7: Create the kafka openssl config 

```bash
      my-kafka-server 192.0.2.78> cat /opt/kafka/config/openssl.cnf 
      [ subject_alt_name ]
```

### STEP 8: Create the logs directory:

```bash
      my-kafka-server> mkdir /opt/kafka/logs
[...]

PR #130, opened 2020-10-20
Initial commit of kafka server configuration

jvlcek:
This PR will automate the steps required to configre kafka messaging service

Fixes https://github.com/ManageIQ/manageiq-appliance_console/issues/125


jvlcek:
@miq-bot add_label wip

lib/manageiq/appliance_console/message_server_configuration.rb
```diff
@@ -0,0 +1,247 @@
...
+
+    attr_reader :hostname, :username, :password, :config_dir, :sample_config_dir,
+      :jaas_config_path, :openssl_config_path, 
+      :log4_properties_path, :log4_properties_sample_path, :client_properties_path,
+      :server_properties_path, :server_properties_orig_path,
+      :keystore_dir, :truststore_path, :keystore_path,
+      :messaging_yaml_sample_path, :messaging_yaml_path
+
+    def initialize(options = {})
+      puts("JJV 0.0 #{File.basename(__FILE__)} / #{__method__}")
```
Fryguy, line 19: Just pinging this to be removed later.

lib/manageiq/appliance_console/message_server_configuration.rb
```diff
@@ -0,0 +1,247 @@
...
+      show_parameters
+      return false unless agree("\nProceed? (Y/N): ")
+      return false unless valid_parameters?(@hostname)
+      true
+    end
+
+    def activate
+      puts("JJV 0.0 #{File.basename(__FILE__)} / #{__method__}")
+      begin
+        configure_log3               ## STEP 3: Create the log4 configuration file
```
Fryguy, line 60: minor but the method is named log3, but the feature is log4?

lib/manageiq/appliance_console/message_server_configuration.rb
```diff
@@ -0,0 +1,247 @@
...
+
+      require 'pry'; binding.pry # JJV
+    end
+
+    def configure_firewall
+      say("#{__method__}".gsub("_", " ").titleize)
+
+      AwesomeSpawn.run!("firewall-cmd --add-port=9092/tcp --permanent") # insecure
+      AwesomeSpawn.run!("firewall-cmd --add-port=9093/tcp --permanent") # secure
+      AwesomeSpawn.run!("firewall-cmd --reload")
```
Fryguy, line 188: Do we do the same thing for postgres?  that is, configure the firewall on the fly?  if so, then :+1:

lib/manageiq/appliance_console/message_server_configuration.rb
```diff
@@ -0,0 +1,247 @@
...
+    def configure_service_properties
+      say("#{__method__}".gsub("_", " ").titleize)
+
+      # Create the .ORIG only on clean environments
+      FileUtils.cp(@server_properties_path, @server_properties_orig_path) unless file_found?(@server_properties_orig_path)
+
+      content = <<~EOS
+
+        ssl.keystore.location="#{keystore_path}"
+        ssl.keystore.password="#{@password}"
```
Fryguy, line 210: Do we have to handle escaping or anything in any of these passwords (here and elsewhere we do interpolation)?  Like, what if someone has a `"` in their password?

Fryguy:
Looking good @jvlcek !

lib/manageiq/appliance_console/message_server_configuration.rb
jvlcek: renamed to create_tools_logconfig

lib/manageiq/appliance_console/message_server_configuration.rb
jvlcek: @Fryguy It doesn't appear that we do firewwall config on the fly for postgres. @bdunne had suggested this approach for kafak. @bdunne Can you please comment?

lib/manageiq/appliance_console/message_server_configuration.rb
jvlcek: @Fryguy I'll investigate. Thank you!

jvlcek:
@miq-bot remove_label wip

@Fryguy and @bdunne Please review.

jvlcek:
Regarding codeclimate and hakiri failures: @Fryguy has told me:

"you can ignore codeclimate...it's been having issues on that repo and I can't turn it off
same with hakiri"

bin/appliance_console
```diff
@@ -320,6 +320,33 @@ Static Network Configuration
         end
         press_any_key
 
+      when I18n.t("advanced_settings.message_server")
```
bdunne, line 323: I think these should live under https://github.com/ManageIQ/manageiq-appliance_console/pull/130/files#diff-e8b26cb149425ac9294176fd67aa5744bcd030ff9ad96fe836bbdfa50607ce0dR449 with the existing external messaging configuration.

lib/manageiq-appliance_console.rb
```diff
@@ -39,6 +39,9 @@ def self.logger=(logger)
 require 'manageiq/appliance_console/external_auth_options'
 require 'manageiq/appliance_console/external_database_configuration'
 require 'manageiq/appliance_console/external_httpd_authentication'
+require 'manageiq/appliance_console/message_configuration'
```
bdunne, line 42: Sort alphabetically?

bdunne: Can this be merged into https://github.com/ManageIQ/manageiq-appliance_console/pull/130/files#diff-7a5a900c5a6ca9e861d311795135eaa7e7556074b5dc71399e90e12a7910b946R49 ?

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -0,0 +1,134 @@
...
+        @username, @password = nil
+
+        @username = options[:username] || "admin"
+        @password = options[:password]
+
+        @miq_config_dir_path               = Pathname.new(MIQ_CONFIG_DIR)
+        @config_dir_path                   = Pathname.new(CONFIG_DIR)
+        @sample_config_dir_path            = Pathname.new(SAMPLE_CONFIG_DIR)
+
+        @tools_log4_properties_path        = @config_dir_path.join("", "tools-log4j.properties")
```
bdunne, line 30: Why the empty quotes on these?

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -0,0 +1,134 @@
...
+        @keystore_path                     = @keystore_dir_path.join("", "keystore.jks")
+
+        @messaging_yaml_sample_path        = @miq_config_dir_path.join("", "messaging.kafka.yml")
+        @messaging_yaml_path               = @miq_config_dir_path.join("", "messaging.yml")
+      end
+
+      def create_tools_log_config
+        say(__method__.to_s.tr("_", " ").titleize)
+
+        FileUtils.cp(@tools_log4_properties_sample_path, @tools_log4_properties_path) unless file_found?(@tools_log4_properties_path)
```
bdunne, line 45: You shouldn't need the `@` here since it's an `attr_reader` (same throughout the PR).

bdunne: What do you think of `tools_log4_properties_path.exist?` (Pathname#exist?) rather than a new `file_found?` method?

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -0,0 +1,134 @@
...
+            password=#{@password} ;
+        CLIENT_PROPERTIES
+
+        File.write(@client_properties_path, content) unless file_found?(@client_properties_path)
+      end
+
+      def configure_messaging_yaml
+        say(__method__.to_s.tr("_", " ").titleize)
+
+        return true if file_found?(@messaging_yaml_path)
```
bdunne, line 68: Do we need to return `true` here?  It doesn't look like it's used anywhere.  Also, if we don't return `true` here, the method will return an integer from `File.write`

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -0,0 +1,134 @@
...
+
+        messaging_yaml = YAML.load_file(@messaging_yaml_sample_path)
+
+        messaging_yaml["production"]["hostname"] = @server_hostname
+        messaging_yaml["production"]["port"] = 9093
+        messaging_yaml["production"]["username"] = @username
+        messaging_yaml["production"]["password"] = @password
+
+        File.write(@messaging_yaml_path, messaging_yaml.to_yaml) unless file_found?(@messaging_yaml_path)
```
bdunne, line 77: The `unless` shouldn't be needed since we checked a few lines above.

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -0,0 +1,134 @@
...
+      end
+
+      def remove_installed_files
+        say(__method__.to_s.tr("_", " ").titleize)
+
+        @installed_files.each { |f| FileUtils.rm_rf(f) if File.exist?(f) }
```
bdunne, line 83: I don't think we need to check `if File.exist?` here

lib/manageiq/appliance_console/message_configuration_server.rb
```diff
@@ -0,0 +1,225 @@
...
+
+        FileUtils.mkdir_p(LOGS_DIR)
+        FileUtils.chmod(0o755, LOGS_DIR)
+        FileUtils.chown("kafka", "kafka", LOGS_DIR)
+      end
+
+      def configure_firewall
+        say(__method__.to_s.tr("_", " ").titleize)
+
+        AwesomeSpawn.run!("firewall-cmd --add-port=9092/tcp --permanent") # insecure
```
bdunne, line 122: Since we're going through the trouble of setting up SSL, I don't think we should listen on 9092 for insecure connections.

bdunne (changes requested):

lib/manageiq/appliance_console/message_configuration.rb
Fryguy: file_found? also logs a message (not saying it should, just that's why there's a separate method)

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -0,0 +1,134 @@
...
+        say("\tWARNING: #{path} already exists. Taking no action.")
+        true
+      end
+
+      def file_contains?(path, content)
+        return false unless File.exist?(path)
+
+        content.split("\n").each do |l|
+          l.gsub!("/", "\\/")
+          l.gsub!(/password=.*$/, "password=") # Remove the passord as it can have special characters that grep can not match.
```
Fryguy, line 110: Typo in comment - passord -> password

lib/manageiq/appliance_console/message_configuration_server.rb
```diff
@@ -0,0 +1,225 @@
...
+      def configure_keystore
+        say(__method__.to_s.tr("_", " ").titleize)
+
+        return true if file_found?(@keystore_dir_path)
+
+        FileUtils.mkdir_p(@keystore_dir_path)
+        FileUtils.chmod(0o755, @keystore_dir_path)
+
+        # Generte a Java keystore and key pair, creating keystore.jks
+        AwesomeSpawn.run!("keytool -keystore #{@keystore_path} -alias localhost -validity 10000 -genkey -keyalg RSA -storepass #{@password} -keypass #{@password} -dname \"cn=#{@server_hostname}\" ")
```
Fryguy, line 136: AwesomeSpawn should be used to handle the individual parameters to ensure they are properly escaped.  For example, if someone puts a `|` character in their password, the current way is broken.  Prefer:

```suggestion
        AwesomeSpawn.run!("keytool", :params => {"-keystore" => @keystore_path, "-alias" => "localhost", "-validity" => 10000, "-genkey" => nil, "-keyalg" => "RSA", "-storepass" => @password, "-keypass" => @password, "-dname" => "cn=#{@server_hostname}")
```

lib/manageiq/appliance_console/message_configuration_server.rb
```diff
@@ -0,0 +1,225 @@
...
+
+        # Use openssl to create a new CA cert, creating ca-cert and ca-key
+        AwesomeSpawn.run!("PASSWORD='#{@password}' openssl req -new -x509 -keyout #{@ca_key_path} -out #{@ca_cert_path} -days 10000 -passout 'env:PASSWORD' -subj '/CN=something'")
```
Fryguy, line 139: And ENV can also be parameterized

```suggestion
        AwesomeSpawn.run!("openssl", :env => {"PASSWORD" => @password}, :params => ["req", "-new", "-x509", {"-keyout" => @ca_key_path, "-out" => @ca_cert_path, "-days" => 10000, "-passout" => "env:PASSWORD", "-subj" => "/CN=something"}])
```

"

bin/appliance_console
bdunne: Better link... https://github.com/ManageIQ/manageiq-appliance_console/blob/481d9b37c73a44471b13ce5c5d7a8fce00b7429d/bin/appliance_console#L449

lib/manageiq-appliance_console.rb
jvlcek: @bdunne Sorry that link is not clear. Where are you suggesting I merge this require into?

bdunne: https://github.com/ManageIQ/manageiq-appliance_console/blob/481d9b37c73a44471b13ce5c5d7a8fce00b7429d/lib/manageiq-appliance_console.rb#L49

bin/appliance_console
jvlcek: @bdunne Thank you. I will rearrange the menu items and group the message updates together with the current messaging prompts.

lib/manageiq-appliance_console.rb
jvlcek: @bdunne Alphabetize requires 👍 As discussed combining messaging_configuration with message_configure_client will be done in the future.

lib/manageiq/appliance_console/message_configuration.rb
jvlcek: @bdunne Good catch! Opps! Fixed by commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/dcc87dd8cf25254869dcdd0724dd44e491c7691c

lib/manageiq/appliance_console/message_configuration.rb
jvlcek: Correct, @Fryguy the _file_found?_ method also logs the message letting  the user know the file already exists.
@bdunne I will remove the unnecessary `@`. I just like to leave them so when reading the code the scope of the variable is immediately  evident. But if it is more acceptable style I',m  OK with removing them.

Updated by commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/a1a2b4a5cf7ab94cfd249c654ae45ce324bdc34f

lib/manageiq/appliance_console/message_configuration.rb
jvlcek: @bdunne Good point. Fixed by commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/a115b2753a13e4ae3bd1b8fe0769aa12bcc96fcf

lib/manageiq/appliance_console/message_configuration.rb
jvlcek: @bdunne Good catch! Fixed by commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/5dcf698752f575400a690b9056d135ea22e7ec3b

lib/manageiq/appliance_console/message_configuration.rb
jvlcek: Well it seems not. I knew FIleUtils.rm produces a traceback if the file does not exist. I didn't realize FileUtils.rm_rf() does not. @bdunne Once again **Good Catch!** 

Fixed by commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/327bca8ea1fab36dee03c3348a904d4af2ff200b

lib/manageiq/appliance_console/message_configuration_server.rb
jvlcek: @bdunne OK. I had meant to ask you opinion on this. 

Addressed by commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/834642c3ea9207676d43da07cf4eff73f94524fa

lib/manageiq/appliance_console/message_configuration.rb
jvlcek: @Fryguy Good catch! I ran spell checker. I wonder why it didn't flag this.

lib/manageiq/appliance_console/message_configuration_server.rb
jvlcek: @Fryguy Thank you for you input. I've addressed this and the other AwesomeSpawn issue with commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/37575cb580cab9d0404b9640c9bad22706ae28e8

jvlcek:
@bdunne and @Fryguy I believe I have addressed all the issues you have identified, so this PR should be ready to merge.

One issue @bdunne and I had discussed, which he agreed could be addressed in a follow on PR, will be tracked in [Issue 135](https://github.com/ManageIQ/manageiq-appliance_console/issues/135)

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -0,0 +1,134 @@
...
+        say(__method__.to_s.tr("_", " ").titleize)
+
+        return if file_found?(messaging_yaml_path)
+
+        messaging_yaml = YAML.load_file(messaging_yaml_sample_path)
+
+        messaging_yaml["production"]["hostname"] = server_hostname
+        messaging_yaml["production"]["port"] = 9093
+        messaging_yaml["production"]["username"] = username
+        messaging_yaml["production"]["password"] = password
```
jvlcek, line 75: In looking at https://github.com/ManageIQ/manageiq-appliance_console/blob/master/lib/manageiq/appliance_console/messaging_configuration.rb#L85 the password should be encrypted here too.

jvlcek: This is addressed by commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/86304c6da9c13f09867e41c9548ec25fb0108dd8

Fryguy (approved):

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -0,0 +1,134 @@
...
+                  :tools_log4_properties_path, :tools_log4_properties_sample_path,
+                  :client_properties_path,
+                  :keystore_dir_path, :truststore_path, :keystore_path,
+                  :messaging_yaml_sample_path, :messaging_yaml_path
+
+      BASE_DIR                          = "/opt/kafka".freeze
+      LOGS_DIR                          = "#{BASE_DIR}/logs".freeze
+      CONFIG_DIR                        = "#{BASE_DIR}/config".freeze
+      SAMPLE_CONFIG_DIR                 = "#{BASE_DIR}/config-sample".freeze
+      MIQ_CONFIG_DIR                    = "/var/www/miq/vmdb/config".freeze
```
bdunne, line 18: Should this use `ManageIQ::ApplianceConsole::RAILS_ROOT.join("config")`?

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -0,0 +1,134 @@
...
+
+      def initialize(options = {})
+        @username, @password = nil
```
bdunne, line 21: Is this needed if they're set on the next two lines?

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -0,0 +1,134 @@
...
+        @keystore_path                     = keystore_dir_path.join("keystore.jks")
+
+        @messaging_yaml_sample_path        = miq_config_dir_path.join("messaging.kafka.yml")
+        @messaging_yaml_path               = miq_config_dir_path.join("messaging.yml")
+      end
+
+      def create_tools_log_config
+        say(__method__.to_s.tr("_", " ").titleize)
+
+        FileUtils.cp(tools_log4_properties_sample_path, tools_log4_properties_path) unless file_found?(tools_log4_properties_path)
```
bdunne, line 45: Did we miss this in the RPM spec?  If so, I'd prefer to fix it there.

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -0,0 +1,134 @@
...
+          ssl.truststore.password=#{password}
+
+          sasl.mechanism=PLAIN
+          security.protocol=SASL_SSL
+          sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required \\
+            username=#{username} \\
+            password=#{password} ;
+        CLIENT_PROPERTIES
+
+        File.write(client_properties_path, content) unless file_found?(client_properties_path)
```
bdunne, line 62: Should we `return unless ...` earlier to avoid creating the heredoc or should we always write the file in case the values changed?

lib/manageiq/appliance_console/message_configuration_server.rb
```diff
@@ -0,0 +1,224 @@
...
+        @ca_key_path                       = keystore_dir_path.join("ca-key")
+        @cert_file_path                    = keystore_dir_path.join("cert-file")
+        @cert_signed_path                  = keystore_dir_path.join("cert-signed")
+
+        @installed_files = [tools_log4_properties_path, jaas_config_path,
+                            client_properties_path, server_properties_path,
+                            messaging_yaml_path, LOGS_DIR, keystore_dir_path]
+      end
+
+      def ask_questions
```
bdunne, line 33: Is there a difference between this method and https://github.com/ManageIQ/manageiq-appliance_console/pull/130/files#diff-b896847105d5f7861df9e75f19292936e53103942ce4a0187fa37cafd3671839R25 ? Should they be merged into the base class?

lib/manageiq/appliance_console/message_configuration_client.rb
```diff
@@ -0,0 +1,118 @@
...
+
+module ManageIQ
+  module ApplianceConsole
+    class MessageClientConfiguration < MessageConfiguration
+      attr_reader :server_hostname, :server_username, :server_password, :installed_files
+
+      def initialize(options = {})
+        super(options)
+
+        @server_hostname, @server_username, @server_password = nil
```
bdunne, line 15: Is this needed since they're defined below?

lib/manageiq/appliance_console/message_configuration.rb
jvlcek: Good point. I didn't even know that existed.
Addressed by commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/9bb79759c88dddb951dad05049193da90aaacbcc

lib/manageiq/appliance_console/message_configuration.rb
jvlcek: @bdunne Good catch.
Addressed by commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/2709f84a43f067589eb5cbca6c02e475d91a4307

lib/manageiq/appliance_console/message_configuration_client.rb
```diff
@@ -0,0 +1,118 @@
...
+        end
+
+        File.exist?(truststore_path)
+      rescue => e
+        say("Failed to fetch server truststore: #{e.message}")
+        false
+      end
+
+      def installation_valid?
+        return true if LinuxAdmin::Rpm.list_installed.key?("kafka")
```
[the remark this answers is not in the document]
bdunne: I un-WIP'd https://github.com/ManageIQ/manageiq-rpm_build/pull/91

lib/manageiq/appliance_console/message_configuration_client.rb
```diff
@@ -0,0 +1,118 @@
...
+
+        say("\nAppliance Installation is not valid for enabling Message\n")
+        false
+      end
+
+      def already_configured?
+        installed_file_found = false
+        installed_files.each do |f|
```
bdunne, line 104: Could be simplified with `installed_files.any? ...`

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -0,0 +1,134 @@
...
+    class MessageConfiguration
+      attr_reader :username, :password,
+                  :miq_config_dir_path, :config_dir_path, :sample_config_dir_path,
+                  :tools_log4_properties_path, :tools_log4_properties_sample_path,
+                  :client_properties_path,
+                  :keystore_dir_path, :truststore_path, :keystore_path,
+                  :messaging_yaml_sample_path, :messaging_yaml_path
+
+      BASE_DIR                          = "/opt/kafka".freeze
+      LOGS_DIR                          = "#{BASE_DIR}/logs".freeze
```
bdunne, line 15: Do you know what's put here?  If it's server logs, we should probably move that to `/var/log`, but if it's message data that they're calling "logs" it's fine to stay here.

lib/manageiq/appliance_console/message_configuration_client.rb
```diff
@@ -0,0 +1,118 @@
...
+      end
+
+      private
+
+      def fetch_truststore_from_server
+        say(__method__.to_s.tr("_", " ").titleize)
+
+        return if file_found?(truststore_path)
+
+        FileUtils.mkdir_p(keystore_dir_path)
```
bdunne, line 82: Should this directory be included in the RPM so that it has the correct permissions and ownership?

bdunne: Or should the two files in here just live in the config directory?

lib/manageiq/appliance_console/message_configuration_server.rb
```diff
@@ -0,0 +1,224 @@
...
+      def configure_keystore
+        say(__method__.to_s.tr("_", " ").titleize)
+
+        return if file_found?(keystore_dir_path)
+
+        FileUtils.mkdir_p(keystore_dir_path)
+        FileUtils.chmod(0o755, keystore_dir_path)
+
+        # Generte a Java keystore and key pair, creating keystore.jks
+        AwesomeSpawn.run!("keytool", :params => {"-keystore" => keystore_path, "-alias" => "localhost", "-validity" => 10_000, "-genkey" => nil, "-keyalg" => "RSA", "-storepass" => password, "-keypass" => password, "-dname" => "cn=#{server_hostname}"})
```
bdunne, line 136: Are we able to put the password in an ENV var to help protect it?

lib/manageiq/appliance_console/message_configuration.rb
jvlcek: @bdunne Good idea.
Addressed by: https://github.com/ManageIQ/manageiq-rpm_build/pull/122
and commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/8545e96290fa0369a7693fb7f2f8cfb3824b49a5

lib/manageiq/appliance_console/message_configuration.rb
jvlcek: It seems to be a balance of one extra line of code in the source or creating the heredoc.
I'll avoid the heredoc.
Addressed by commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/48209af0a4f77eff5d9286b3617d0dea94fad40d

lib/manageiq/appliance_console/message_configuration_server.rb
jvlcek: Good catch! Thanks!

Addressed by commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/13470ad9c2c671cea76655a630e7c0752dee8e4e

lib/manageiq/appliance_console/message_configuration_client.rb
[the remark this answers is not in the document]
jvlcek: Ah, OK. Good idea.
Addressed by commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/c3df35e082ad7102e1f62be15cb5b08a46be334f

lib/manageiq/appliance_console/message_configuration_server.rb
```diff
@@ -0,0 +1,224 @@
...
+      end
+
+      def installation_valid?
+        return true if LinuxAdmin::Rpm.list_installed.key?("kafka")
+
+        say("\nAppliance Installation is not valid for enabling Message\n")
+        false
+      end
+
+      def already_configured?
```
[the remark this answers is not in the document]
jvlcek: Good catch. I'm sorry but I fixed this with the same commit where I removed checking for the kafka RPM.

Addressed in commit https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/c3df35e082ad7102e1f62be15cb5b08a46be334f

lib/manageiq/appliance_console/message_configuration_client.rb
jvlcek: Good point. I also fixed my incorrect fix of unnecessary initialization to nil in commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/9406e5bfb02693d7a32e0f853f6ef9a3a889a21b

lib/manageiq/appliance_console/message_configuration_client.rb
jvlcek: @bdunne I'm not sure how because I want to `say("Installed file #{f} found.")` to list every file found. It seems `installed_files.any?` returns `true` the first time the block returns `true`.

lib/manageiq/appliance_console/message_configuration.rb
jvlcek: @bdunne 

It seems to be server logs. I'm OK with reconfiguring Kafka to write to `/var/log` but it seems kafka's logging configuration is a bit complex so I if we decide these logs really should be relocated I'd like to tackle that in a follow on PR.
Are you OK with that?

lib/manageiq/appliance_console/message_configuration_client.rb
jvlcek: @bdunne Good idea to create this directory in the RPM.

Updated the RPM spec to do just that in commit: https://github.com/ManageIQ/manageiq-rpm_build/pull/122/commits/a8b119569777798e03e28a4a9be87d416bca7a46

Updated here in commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/130/commits/4d74dd3ab8fa320f25d0fcf40cc039a5fc71cbe9

lib/manageiq/appliance_console/message_configuration_server.rb
```diff
@@ -0,0 +1,224 @@
...
+        AwesomeSpawn.run!("keytool", :params => {"-keystore" => keystore_path, "-alias" => "CARoot", "-import" => nil, "-file" => ca_cert_path, "-storepass" => password, "-noprompt" => nil})
+
+        # Import a signed primary certificate to an existing Java keystore, updating keystore.jks
+        AwesomeSpawn.run!("keytool", :params => {"-keystore" => keystore_path, "-alias" => "localhost", "-import" => nil, "-file" => cert_signed_path, "-storepass" => password, "-noprompt" => nil})
+      end
+
+      def create_server_properties
+        say(__method__.to_s.tr("_", " ").titleize)
+
+        content = <<~SERVER_PROPERTIES
```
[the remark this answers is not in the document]
jvlcek: @bdunne Yes there are roughly 20 name=value pairs already in the server.properties file. I think doing it this way could be safer than hardcoding what is delivered in the file because future versions of kafka could be delivered with different content. Then we would have to update the content here too.
I propose we leave this as is.

lib/manageiq/appliance_console/message_configuration_server.rb
jvlcek: @bdunne I spent a fair amount of time exploring and experimenting with this and it does not see possible to pass passwords to keytool in an ENV var.

lib/manageiq/appliance_console/message_configuration_client.rb
bdunne: Ok

lib/manageiq/appliance_console/message_configuration.rb
bdunne: Ok, followup sounds good.

bdunne:
Followup in https://github.com/ManageIQ/manageiq-appliance_console/issues/135

PR #132, opened 2020-10-26
Fix .rubocop_local.yml

Fryguy:
Fix .rubocop_local.yml


bdunne:
Should we empty the file instead?

Fryguy:
For now let's merge since I blaster outered this, but yes, we can do that in a follow up.

PR #134, opened 2020-11-10
Remove rbnacl-libsodium, will add rbnacl-devel RPM

jvlcek:
see: https://github.com/RubyCrypto/rbnacl-libsodium#discontinued

@Fryguy, @agrare, @gtanzillo and @simaishi Need to add the rbnacl-devel RPM in PR:
https://github.com/ManageIQ/manageiq-rpm_build/pull/117

### DIFF
--- a/manageiq-appliance_console.gemspec
+++ b/manageiq-appliance_console.gemspec
@@ -33,7 +33,6 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "optimist",                "~> 3.0"
   spec.add_runtime_dependency "pg"
   spec.add_runtime_dependency "rbnacl",                  ">= 3.2", "< 5.0"
-  spec.add_runtime_dependency "rbnacl-libsodium"
 
   spec.add_development_dependency "bundler"
   spec.add_development_dependency "codeclimate-test-reporter", "~> 1.0.0"


Fryguy (approved):
LGTM - will wait for the other PRs and merge them all together.

Fryguy:
Not sure how to release this...I'm thinking we should just a major version?  We don't want this accidentally pulled in to a system that doesn't have the rbnacl-devel package installed?

Fryguy (changes requested):
@jvlcek If we remove this it will break the rpm build because we play some games with the on-disk assets post-bundle-install.  I think you will have to remove the following

https://github.com/ManageIQ/manageiq-rpm_build/blob/0f8f85208c656989608a8493d5aac0439ce12242/lib/manageiq/rpm_build/generate_gemset.rb#L82-L87

https://github.com/ManageIQ/manageiq-rpm_build/blob/0f8f85208c656989608a8493d5aac0439ce12242/lib/manageiq/rpm_build/generate_gemset.rb#L173-L176


cc @simaishi

jvlcek:
> @jvlcek If we remove this it will break the rpm build because we play some games with the on-disk assets post-bundle-install. I think you will have to remove the following
> 
> https://github.com/ManageIQ/manageiq-rpm_build/blob/0f8f85208c656989608a8493d5aac0439ce12242/lib/manageiq/rpm_build/generate_gemset.rb#L82-L87
> 
> https://github.com/ManageIQ/manageiq-rpm_build/blob/0f8f85208c656989608a8493d5aac0439ce12242/lib/manageiq/rpm_build/generate_gemset.rb#L173-L176
> 
> cc @simaishi

Please see: https://github.com/ManageIQ/manageiq-rpm_build/pull/117/commits/c653fd640865d4053ed6171bf2d71788a932c83d

Issue #136, opened 2020-11-20
Toggle the Settings.prototype.messaging_type  for Kafka support


jvlcek:
In addition to establishing the Kafka configuration the appliance console will also have to set the messaging_type Settings to _kafka_ when activating kafka.

For example:
`Settings.prototype.messaging_type = kafka`

And back to _miq_queue_ when deactivating kafka

For example:
`Settings.prototype.messaging_type = miq_queue`

jvlcek:
See also:
Once client configuration is complete on a server, switch https://github.com/ManageIQ/manageiq/blob/2f8ae2345eb25b16ad87f7dc124b4fe68a610bdb/config/settings.yml#L936 to kafka.

jvlcek:
@miq-bot add_label enhancement
@miq-bot assign @jvlcek

PR #137, opened 2020-11-20
Toggle Settings.prototype.messaging_type for Kafka support

jvlcek:
This PR will add support for setting the messaging_type Settings to _kafka_ when activating kafka and back to _miq_queue_ when deactivating kafka

Fixes https://github.com/ManageIQ/manageiq-appliance_console/issues/136

### DIFF
--- a/lib/manageiq/appliance_console/message_configuration_client.rb
+++ b/lib/manageiq/appliance_console/message_configuration_client.rb
@@ -25,6 +25,7 @@ module ManageIQ
           create_client_properties          # Create the client.properties configuration fle
           fetch_truststore_from_server      # Fetch the Java Keystore from the Kafka Server
           configure_messaging_type("kafka") # Settings.prototype.messaging_type = 'kafka'
+          restart_evmserverd
         rescue AwesomeSpawn::CommandResultError => e
           say(e.result.output)
           say(e.result.error)
@@ -36,12 +37,6 @@ module ManageIQ
           return false
         end
         true
-      end
-
-      def post_activation
-        say("Restart evmserverd if it is running...")
-        evmserverd_service = LinuxAdmin::Service.new("evmserverd")
-        evmserverd_service.restart if evmserverd_service.running?
       end
 
       def ask_for_parameters
@@ -81,8 +76,9 @@ module ManageIQ
       end
 
       def deactivate
-        remove_installed_files
         configure_messaging_type("miq_queue") # Settings.prototype.messaging_type = 'miq_queue'
+        restart_evmserverd
+        remove_installed_files
       end
     end
   end
--- a/lib/manageiq/appliance_console/message_configuration_client.rb
+++ b/lib/manageiq/appliance_console/message_configuration_client.rb
@@ -21,9 +21,10 @@ module ManageIQ
 
       def activate
         begin
-          configure_messaging_yaml     # Set up the local message client in case EVM is actually running on this, Message Server
-          create_client_properties     # Create the client.properties configuration fle
-          fetch_truststore_from_server # Fetch the Java Keystore from the Kafka Server
+          configure_messaging_yaml          # Set up the local message client in case EVM is actually running on this, Message Server
+          create_client_properties          # Create the client.properties configuration fle
+          fetch_truststore_from_server      # Fetch the Java Keystore from the Kafka Server
+          configure_messaging_type("kafka") # Settings.prototype.messaging_type = 'kafka'
         rescue AwesomeSpawn::CommandResultError => e
           say(e.result.output)
           say(e.result.error)
@@ -75,6 +76,7 @@ module ManageIQ
 
       def deactivate
         remove_installed_files
+        configure_messaging_type("miq_queue") # Settings.prototype.messaging_type = 'miq_queue'
       end
     end
   end
--- a/lib/manageiq/appliance_console/message_configuration_client.rb
+++ b/lib/manageiq/appliance_console/message_configuration_client.rb
@@ -38,6 +38,12 @@ module ManageIQ
         true
       end
 
+      def post_activation
+        say("Restart evmserverd if it is running...")
+        evmserverd_service = LinuxAdmin::Service.new("evmserverd")
+        evmserverd_service.restart if evmserverd_service.running?
+      end
+
       def ask_for_parameters
         say("\nMessage Client Parameters:\n\n")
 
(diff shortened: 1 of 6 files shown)


jvlcek:
@miq-bot add_label wip
@miq-bot assign @bdunne

jvlcek:
This PR depends on https://github.com/ManageIQ/manageiq/pull/20837

jvlcek:
Todo: Add specs

lib/manageiq/appliance_console/message_configuration_client.rb
```diff
@@ -37,6 +38,12 @@ def activate
         true
       end
 
+      def post_activation
+        say("Restart evmserverd if it is running...")
+        evmserverd_service = LinuxAdmin::Service.new("evmserverd")
+        evmserverd_service.restart if evmserverd_service.running?
```
bdunne, line 44: Do we need to restart the server?  It should get picked up automatically, right?

agrare: We cache the [`MiqQueue.messaging_client`](https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_queue.rb#L39-L58) so if we switched from kafka->miq-queue, artemis->miq-queue, kafka->artemis, or artemis->kafka we would have potentially stale connections cached.  We could handle this a different way though e.g. check if the messaging_type is mismatched with an existing messaging_client.

bdunne: Maybe a job for the config activator?

lib/manageiq/appliance_console/message_configuration_client.rb
```diff
@@ -75,6 +82,7 @@ def fetch_truststore_from_server
 
       def deactivate
         remove_installed_files
+        configure_messaging_type("miq_queue") # Settings.prototype.messaging_type = 'miq_queue'
```
bdunne, line 85: Should we change the messaging type first, restart the server then remove the installed files?

jvlcek: Good point.
How about this:
```
stop_evmserverd # New method
remove_installed_files
configuremessaging_type("miq_queue")
restart evmserverd
```

lib/manageiq/appliance_console/message_configuration_client.rb
jvlcek: @bdunne Yes the restart of evmserverd is needed. The message_type setting is not picked up automatically.

lib/manageiq/appliance_console/message_configuration_client.rb
jvlcek: Addressed in commit https://github.com/ManageIQ/manageiq-appliance_console/pull/137/commits/42884bb4877b494dfde0e2a06e3f25dc139a17d2

jvlcek:
@miq-bot remove_label wip

jvlcek:
Removing WIP as, with @agrare help, I have tested with a ManageIQ appliance. There are some changes that need to be done on the appliance side as it had not be tested with rdkafka and SSL. Those PRs are on the way.

lib/manageiq/appliance_console/message_configuration_client.rb
jvlcek: @bdunne I think cleaning up the potentially stale cached connections is beyond the scope of this PR, and as you mention could be addressed by the config activator in a separate PR.

spec/message_configuration_client_spec.rb
```diff
@@ -193,4 +196,18 @@
       expect(subject.send(:fetch_truststore_from_server)).to be_nil
     end
   end
+
+  describe "#restart_evmserverd" do
```
bdunne, line 200: Should these examples have an expectation on `restart` being called?

spec/message_configuration_server_spec.rb
```diff
@@ -283,4 +286,22 @@
       expect(subject.send(:configure_messaging_yaml)).to be_nil
     end
   end
+
+  describe "#post_activation" do
```
bdunne, line 290: Same here, should these examples expect `start` commands?

spec/message_configuration_client_spec.rb
jvlcek: @bdunne Agreed but I think the do. Shouldn't this do it:

209       expect(LinuxAdmin::Service).to receive(:new).with("evmserverd").and_return(double(@spec_name, :running? => false))

spec/message_configuration_server_spec.rb
jvlcek: @bdunne shouldn't this do it?

295       expect(LinuxAdmin::Service).to receive(:new).with("evmserverd").and_return(double(@spec_name, :running? => true, :restart => nil))

spec/message_configuration_client_spec.rb
bdunne: That just means that calling `restart` on that `double` will return `nil`, right?

bdunne: I don't think it set up any expectation that `restart` gets called

jvlcek: I get it now. Thank you! Will fix.

spec/message_configuration_client_spec.rb
```diff
@@ -193,4 +196,22 @@
...
+    it "restarts evmserverd if it is running" do
+      expect(subject).to receive(:say)
+      evmserverd = double(@spec_name, :running? => true)
```
bdunne, line 203:
```suggestion
      evmserverd = LinuxAdmin::Service.new("evmserverd")
      expect(evmserverd).to receive(:running?).and_return(true)
```

spec/message_configuration_client_spec.rb
```diff
@@ -193,4 +196,22 @@
...
+      expect(evmserverd).to receive(:restart)
+      expect(LinuxAdmin::Service).to receive(:new).with("evmserverd").and_return(evmserverd)
+      expect(subject.send(:restart_evmserverd)).to be_nil
+    end
+
+    it "does not restart evmserverd if it is not running" do
+      expect(subject).to receive(:say)
+      evmserverd = double(@spec_name, :running? => false)
```
bdunne, line 211:
```suggestion
      evmserverd = LinuxAdmin::Service.new("evmserverd")
      expect(evmserverd).to receive(:running?).and_return(false)
```

jvlcek: @bdunne Thank you for the help. I believe it is good now.

jvlcek:
@kbrock Thank you for the help!

Issue #141, opened 2020-12-02
When configuring the kafka client disable the kafka server


jvlcek:
To reduce the overhead of the kafka server, when configuring the kafka client on an appliance disable the kafka server

PR #154, opened 2021-03-09
When configuring the kafka client disable the server

jvlcek:
Fixes https://github.com/ManageIQ/manageiq-appliance_console/issues/141

The overhead imposed by running the kafka server is unnecessary when only the kafka client is desired.

The best way to ensure resources used by the kafka server are cleaned up is to invoke the kafka servers `#unconfigure` method. Running the `#unconfigure` method is only necessary if the kafka or zookeeper services are not running. It could be argued that the new method `#configured?` only need check if the kafka service is running but to be safe I check if either kafka or zooker services are running.

### DIFF
--- a/spec/message_configuration_server_spec.rb
+++ b/spec/message_configuration_server_spec.rb
@@ -294,5 +294,18 @@ describe ManageIQ::ApplianceConsole::MessageServerConfiguration do
 
       expect(described_class.configured?).to be_truthy
     end
+
+    it "returns false if neither the zookeeper service or the kafka service are running" do
+      kafka = LinuxAdmin::Service.new("kafka")
+      expect(kafka).to receive(:running?).and_return(false)
+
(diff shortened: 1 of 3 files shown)


jvlcek:

jvlcek:
@miq-bot assign @bdunne

spec/message_configuration_server_spec.rb
```diff
@@ -272,4 +272,27 @@
       subject.send(:restart_services)
     end
   end
+
+  describe "#configured?" do
```
bdunne, line 276: Can you add a test case for false?

jvlcek: Thank you @bdunne. That would indeed be a good idea!

Addressed by commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/154/commits/13a4c579bc5e970afbd502c5a8194e5a3b10f391

PR #153, opened 2021-03-05
[WIP] Inject postgres admin into appliance console

NickLaMuro:
Part of a effort to move database backups in ManageIQ to be only an appliance console concern.

Currently, the only places that `PostgresAdmin` is used is for database backups:

https://github.com/search?p=1&q=org%3AManageIQ+PostgresAdmin&type=Code

The code on the `manageiq` side only uses it for database backups (mostly in `lib/evm_database_ops.rb`, and the other place is here.  Since this will be the only place using it otherwise, this ports it into from `manageiq-gems-pending` into this repo, and moves it into a more reasonable location for this repo.

Still a work in progress, but arguably could be merged, and the removal bit in `manageiq-gems-pending` can be saved from later.

Side Benefit:  This might be the only requirement this repo has for `manageiq-gems-pending`, so we might be able to remove that as well.  We will see.

Process for the extraction
--------------------------

Been working on converting `code-extractor` to use `git-filter-repo`, and the current implementation of that was using a pure `python3` implementation:

https://github.com/juliancheal/code-extractor/compare/master...NickLaMuro:use-python

The `yaml` I used for that was pretty simple:

```yaml
:name: inject_postgres_admin
:upstream: "git@github.com:NickLaMuro/manageiq-gems-pending.git"
:upstream_name: "ManageIQ/manageiq-gems-pending"
:extractions:
  - lib/gems/pending/util/postgres_admin.rb
  - spec/util/postgres_admin_spec.rb
```

After that, there was a few extra commands that I needed to do to make this PR:

```console
$ cd inject_postgres_admin
$ git remote add appliance_console git@github.com:NickLaMuro/manageiq-appliance_console
$ git fetch appliance_console
$ git checkout inject_postgres_admin_into_appliance_console
$ git rebase --rebase-merges=rebase-cousins --root --onto appliance_console/master inject_postgres_admin_into_appliance_console
$ mv lib/gems/pending/util/postgres_admin.rb lib/manageiq/appliance_console/
$ mv spec/util/postgres_admin_spec.rb spec/
$ rm -rf lib/gems spec/util
$ git add -A
$ git commit -m "[PostgresAdmin] Move to lib"
$ git push -u appliance_console inject_postgres_admin_into_appliance_console
```

PR #155, opened 2021-03-10
[Utilities] Add #disk_usage

NickLaMuro:
As one of two things that is still making use of `manageiq-gems-pending` (the other being `PostgresAdmin`), this only a single use method that doesn't need to bring then entirety of `manageiq-gems-pending` along with it in the build.

For `#disk_usage`, porting over the singular method itself, while removing the `linux` vs `mac` differences (only keeping the linux), should be good enough for the use case required by the appliance console.

### DIFF
--- a/lib/manageiq/appliance_console/logfile_configuration.rb
+++ b/lib/manageiq/appliance_console/logfile_configuration.rb
@@ -1,7 +1,7 @@
+require "manageiq/appliance_console/utilities"
 require 'linux_admin'
 require 'pathname'
 require 'fileutils'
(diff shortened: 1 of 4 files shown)


NickLaMuro:
@Fryguy I think this one is actually good to go.  I didn't handle any other case of `MiqSystem.disk_usage` yet, but this should be able to set us up for removing `manageiq-gems-pending` from this repo when `PostgresAdmin` is integrated in https://github.com/ManageIQ/manageiq-appliance_console/pull/153 (still a bit of work to do on that one though)

PR #156, opened 2021-03-15
[.gemspec] Update activesupport/activerecord

NickLaMuro:
Fixes some Hikari warnings (false positives for our use cases here), but also matches the versions found in ManageIQ/manageiq

### DIFF
--- a/manageiq-appliance_console.gemspec
+++ b/manageiq-appliance_console.gemspec
@@ -20,8 +20,8 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_runtime_dependency "activerecord",            ">= 4.2.2"
-  spec.add_runtime_dependency "activesupport",           ">= 4.2.2"
+  spec.add_runtime_dependency "activerecord",            "~> 6.0.3.5"
+  spec.add_runtime_dependency "activesupport",           "~> 6.0.3.5"
   spec.add_runtime_dependency "awesome_spawn",           "~> 1.4"
   spec.add_runtime_dependency "bcrypt",                  "~> 3.1.10"
   spec.add_runtime_dependency "bcrypt_pbkdf",            ">= 1.0", "< 2.0"


NickLaMuro:
@Fryguy thoughts on this one?  Will help allow #153 to be fully "green" instead of being considered a broken build because of some improperly flagged dependencies.

Fryguy:
Lgtm, but it will require a major version update.  Even so, that's ok with me

manageiq-appliance_console.gemspec
```diff
@@ -20,8 +20,8 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_runtime_dependency "activerecord",            ">= 4.2.2"
-  spec.add_runtime_dependency "activesupport",           ">= 4.2.2"
+  spec.add_runtime_dependency "activerecord",            "~> 6.0.3.5"
+  spec.add_runtime_dependency "activesupport",           "~> 6.0.3.5"
```
Fryguy, line 24: Can we change it to allow something like `~> 6.0.3, >= 6.0.3.5`

NickLaMuro: @Fryguy I actually tried that, or at least with this syntax:

```ruby
  spec.add_runtime_dependency "activerecord",            "~> 6.0.0", ">= 6.0.3.5"
  spec.add_runtime_dependency "activesupport",           "~> 6.0.0", ">= 6.0.3.5"
```

And `hakiri` wasn't having it.  Might not properly parse multiple version numbers properly.

I can do possibly `">= 6.0.3.5", "~> 6.0.0"` and maybe that will work, but unsure.  Didn't want to `>= 6.0.3.5` on it's own though.

NickLaMuro: Of note:  I had two other attempts at this before the current result, but they have been rebased out.

Here is what I did try:

### Attempt #1

"Give me the newest 6.0.X.X release of `activerecord`/`activesupport`"

```diff
-  spec.add_runtime_dependency "activerecord",            ">= 4.2.2"
-  spec.add_runtime_dependency "activesupport",           ">= 4.2.2"
+  spec.add_runtime_dependency "activerecord",            "~> 6.0.0"
+  spec.add_runtime_dependency "activesupport",           "~> 6.0.0"
```

### Attempt #2

"Give me the newest 6.0.X.X release of `activerecord`/`activesupport` that is greater than or equal to `6.0.3.5`"

```diff
-  spec.add_runtime_dependency "activerecord",            ">= 4.2.2"
-  spec.add_runtime_dependency "activesupport",           ">= 4.2.2"
  spec.add_runtime_dependency "activerecord",            "~> 6.0.0", ">= 6.0.3.5"
  spec.add_runtime_dependency "activesupport",           "~> 6.0.0", ">= 6.0.3.5"
```

NickLaMuro: > I can do possibly `">= 6.0.3.5", "~> 6.0.0"` and maybe that will work...

Nope, still didn't work.  Going to revert.

Fryguy: Hakiri has always been a pain with these.

Issue #143, opened 2020-12-02
Investigate moving the persistent Kafka queue messages, to a dedicated another volume


jvlcek:
Similar to "Investigate if logs directory should be moved to /var/log #130 (comment)" where are we storing the persistent queue messages, should we have another volume dedicated to it? Also we should look at message retention settings

PR #158, opened 2021-03-20
Move the kafka persistent message storage to a dedicated disk

jvlcek:
Fixes https://github.com/ManageIQ/manageiq-appliance_console/issues/143

This PR allows the administrator to optionally move the kafka persistent message storage to a dedicated disk volume.
Kafka maintains the persistent message storage on the file system under the directory specified by the `log.dirs` parameter as set in the `server.properties` file. Kafka also keeps track of what `log.dirs` was set to. If `log.dirs` is updated kafka will migrate the persistent messages to the new location.

If the mount point for the new dedicated disk volume is the same as the old value for `log.dirs` kafka fails to restart because the old persistent message store is hidden by the new, `empty` mount point. An effective work around for this it to direct the new mount point for the dedicated disk volume to a different directory and to update the `log.dirs` parameter in the `server.properties` file to point to the new directory. This approach has the added benefit that kafka will migrate the old persistent message store to the new location Since we plan to configure kafka on first boot it is highly likely the this automatic migration will be valuable.

### DIFF
--- a/lib/manageiq/appliance_console/message_configuration_server.rb
+++ b/lib/manageiq/appliance_console/message_configuration_server.rb
@@ -11,7 +11,7 @@ module ManageIQ
                   :ca_cert_srl_path, :ca_key_path, :cert_file_path, :cert_signed_path,
                   :keystore_files, :installed_files, :message_persistent_disk
 
-      PERSISTENT_DIRECTORY = Pathname.new("/var/lib/kafka_persistent_data").freeze
+      PERSISTENT_DIRECTORY = Pathname.new("/var/lib/kafka/persistent_data").freeze
       PERSISTENT_NAME = "kafka_messages".freeze
 
       def initialize(options = {})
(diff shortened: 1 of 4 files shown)


jvlcek:
@miq-bot assign @bdunne

@bdunne Here is the next PR for the kafka work.

Please take a look.

lib/manageiq/appliance_console/message_configuration_server.rb
```diff
@@ -9,12 +9,16 @@ class MessageServerConfiguration < MessageConfiguration
       attr_reader :jaas_config_path,
                   :server_properties_path, :server_properties_sample_path,
                   :ca_cert_srl_path, :ca_key_path, :cert_file_path, :cert_signed_path,
-                  :keystore_files, :installed_files
+                  :keystore_files, :installed_files, :message_persistent_disk
+
+      PERSISTENT_DIRECTORY = Pathname.new("/var/lib/kafka_persistent_data").freeze
```
bdunne, line 14: Should this be under `/var/lib/kafka` or a directory owned by the RPM?

jvlcek: Address by commit: https://github.com/ManageIQ/manageiq-appliance_console/pull/158/commits/f7a5672bd6a1928f926c060c9a7aea435c0ca26a

PR #159, opened 2021-03-25
Support  configuring the kafka server with the current IPAddr

jvlcek:
This PR adds a support for a CLI flag that will allow the user to request that the Kafka server be configured with the IP address, without having to explicitly specifying the IP address on the command line.

This is required for the solution to https://github.com/ManageIQ/manageiq-appliance-build/issues/454


jvlcek:
@miq-bot assign @bdunne

@bdunne Here is the next PR for the kafka work.

Please take a look.

bdunne (approved):

PR #160, opened 2021-04-05
Use PostgresAdmin directly for DB backups

NickLaMuro:
Tested using https://github.com/NickLaMuro/manageiq-db_backup-tests to validate the changes.

There are a bunch of "deletes" that will follow this change, but this should not only be merged, but also released as a new version before those changes are merged.  Otherwise the nightly builds of the appliance console will break.

TODO
----

* [x] Need to test against SMB/NFS
  - ~~should work just like local, but currently have to rework connecting to the SMB/NFS shares prior to the tests running (as part of the test suite), and that currently doesn't exist~~ simpler than expected, and there are just a few kinks to work out.  Will post a update to the `manageiq-db_backup-tests` when it is available.
  - https://github.com/NickLaMuro/manageiq-db_backup-tests/pull/1
* [x] Determine if we want to keep file splitting
  - ~~Probably not, but worth mentioning that this is a feature that will no longer be supported~~  Answer:  Nope, burn it.
* [x] Fix specs
  - See https://github.com/ManageIQ/manageiq-appliance_console/pull/160#issuecomment-819835043 for details

Links
-----

* Part of https://github.com/ManageIQ/manageiq/issues/21100


lib/manageiq/appliance_console/database_admin.rb
```diff
@@ -364,15 +176,81 @@ def processing_message
...
+      def restore
+        backup_type = validate_backup_file_type
+
+        if application_connections?
+          message = "Database restore failed. Shut down all evmserverd processes before attempting a database restore"
+          ManageIQ::ApplianceConsole.logger.error(message)
+          raise message
+        end
+
+        MiqRegion.replication_type = :none
```
NickLaMuro, line 210: @Fryguy and/or @jrafanie So most of the methods in "green" here are pulled from `lib/evm_database_ops.rb` in `ManageIQ/manageiq`:

https://github.com/ManageIQ/manageiq/blob/4abf8c6ddd9f1985c1264ca407837423695b1dfd/lib/evm_database_ops.rb#L183-L216

and I was able to convert most of what was there previously to raw `Pg` calls, but this one seems specific.  After taking a look myself, I wasn't able to determine why this line specifically was added, as it seems that it will always be effectively a "no-op" when the method involved is called:

https://github.com/ManageIQ/manageiq/blob/61c77793f44a2453c0908997d4a8c5f79607faa3/app/models/miq_region.rb#L156-L165

But something tells me I am missing something.  Can you enlighten me on what I am missing?

Fryguy: `replication_type` describes whether or not this database is the global side of replication or the remote side of replication.  From a DB backup/restore point of view though I can't see why that would be important.

Fryguy: Oh I think what's happening here is that it's temporarily disabling replication.  Setting replication_type to none would essentially unregister the replication provider.

carbonin: There was a bug where restore couldn't drop the database because the replication backend was holding a connection open I think. There could have very well been more to it than that, but it should all bit in git history/bugzilla somewhere.

Also :wave:

carbonin: https://bugzilla.redhat.com/show_bug.cgi?id=1444993

PR: https://github.com/ManageIQ/manageiq/pull/16942

NickLaMuro: @carbonin Okay, I think I see where my confusion was.

We pass into `.replication_type=` the `desired_type`, but the `if current_type` statements are getting the values from `.replication_type`, which will do some destroy/configure actions if needed, and I assume the methods called in `.replication_type` query against Postgres/PgLogical/stuffIVaguelyKnowAbout

Looks like I might need to look into what those do further, and see if we need to somehow replicate that logic over in the console.

Thanks for the response (and 👋 )

Fryguy: oh hai @carbonin !! 👋

lib/manageiq/appliance_console/postgres_admin.rb
```diff
@@ -320,6 +331,83 @@ class << self
...
+      result = [{"count" => 0}]
+
+      with_pg_connection do |conn|
+        result = conn.exec("SELECT COUNT(pid) FROM pg_stat_activity WHERE application_name LIKE '%MIQ%'")
+      end
+
+      result[0]["count"] > 0
+    end
+
+    private_class_method def self.disable_replication(dbname)
```
NickLaMuro, line 361: I think this will handle all of the necessary concerns that were brought up previously:

https://github.com/ManageIQ/manageiq-appliance_console/pull/160/files/650c62b4dfc42ba519df50c8a8553b5aee848060#diff-4f5a9d143c19dfcadcdb0467ca432e5c0eba904c896862e8af14277978ec1479

I didn't delete the `MiqRegion`, as is currently done in `PglogicalSubscription`, but since this is a restore, that data will be blown away anyway, so we only need the parts that will be used prevent an error due to the HA code preventing a restore (which I think this is all of it... but I will double check).

I condensed a lot of what was in `PglogicalSubscription` to be a single method, so much of what is used here is spread out in the class between `.find`, `.find_all`, `.delete`, `.safe_delete`, and others.  I also removed some of the column id conversion as I didn't think it was necessary, and just referenced it as `sub_id` for the purposes of the `.drop_subscription` portion of the code below.

NickLaMuro: So I did test this on a local appliance using:

https://github.com/NickLaMuro/manageiq-db_backup-tests/pull/1

And it worked just fine, so at least from a non-HA appliance setup, these new changes don't seem to break anything.  Don't really know how to set up a HA appliance to test the scenarios that were present in https://bugzilla.redhat.com/show_bug.cgi?id=1444993 so that is still a unknown, but the base case at least still works.

lib/manageiq/appliance_console/postgres_admin.rb
```diff
@@ -320,6 +331,84 @@ class << self
...
+      result = nil
+
+      with_pg_connection do |conn|
+        query  = "SELECT COUNT(pid) FROM pg_stat_activity"
+        query << " WHERE backend_type = 'client backend'" if backup_type == :basebackup
+        query << " WHERE datname = '#{dbname}'"           if backup_type == :pgdump
+        result = conn.exec(query)
+      end
+
+      result = conn.result[0]["count"].to_i
```
Fryguy, line 398: result is set on line 395, but then it's not used here.  Should this be something like?

```suggestion
      result = result[0]["count"].to_i if result
```

Fryguy: In fact, conn is not available here either, so as-is this can't work.

locales/appliance/en.yml
```diff
@@ -39,22 +39,6 @@ en:
     summary: Summary Information
     quit: Quit
   database_admin:
-    menu_order:
-    - local
-    - nfs
-    - smb
-    - s3
-    - ftp
-    - swift
```
Fryguy, line 48: 😍

lib/manageiq/appliance_console/postgres_admin.rb
```diff
@@ -320,6 +331,84 @@ class << self
...
+      result[0]["count"].to_i > 0
+    end
+
+    private_class_method def self.disable_replication(dbname)
+      require 'pg/logical_replication'
+
+      with_pg_connection do |conn|
+        pglogical = PG::LogicalReplication::Client.new(conn)
+
+        if pglogical.subscriber?
```
Fryguy, line 368: If it makes sense, I'm ok with pushing down any of this logic into the pg/logical_replication gem.  This could also be done in a follow-up

lib/manageiq/appliance_console/postgres_admin.rb
NickLaMuro: Yeah, I fixed this prior to seeing this comment.  There was a decent amount wrong here that I am currently in the process of fixing (hence the current test failures).

NickLaMuro:
So I have determined that there is a failure with the specs here that is based on the order that which they are run in... but have failed to track down what it is...

The `[FIXUP]` commits are currently some of those fixes, but I think one or two more will be necessary for me to track down this bug in the tests.  Probably test specific related at this point, and just based on some of the changes I introduced into `PostgresAdmin` in this PR (which was setting up changes for https://github.com/ManageIQ/manageiq-appliance_console/pull/161 ).

NickLaMuro:
Finally... it passed...

Now I just have to get the PR rebased... but that is a problem for "Monday Nick" to deal with...

lib/manageiq/appliance_console/postgres_admin.rb
NickLaMuro: Just because of the delay I have had trying to get this PR, I am going to punt on doing this now and handle in a follow up.

So we don't loose track of it, I will open up an issue that can be triaged later, and we can determine if it makes sense then to push this into the gem.

NickLaMuro: https://github.com/ManageIQ/manageiq-appliance_console/issues/162

NickLaMuro:
@Fryguy I think this is _**finally**_ ready for your review.  Tests should be passing, and I have rebased/cleaned up the commits.

lib/manageiq/appliance_console/postgres_admin.rb
```diff
@@ -320,6 +329,84 @@ class << self
...
+
+      with_pg_connection do |conn|
+        result = conn.exec("SELECT COUNT(pid) FROM pg_stat_activity WHERE application_name LIKE '%MIQ%'")
+      end
+
+      result[0]["count"].to_i > 0
+    end
+
+    private_class_method def self.disable_replication(dbname)
+      require 'pg/logical_replication'
```
Fryguy, line 361: Do we have to include this gem in the gemspec now?

NickLaMuro: Well, if you searched for it using the `pg/...`, then yeah, it wouldn't show up.  But yes, it is required, but I think the specs wouldn't have passed if I didn't have it. 😉

spec/postgres_admin_spec.rb
```diff
@@ -21,28 +22,29 @@
       end
     end
 
-    context "with a pg_dump file from a pipe" do
-      let(:dbname)    { "pg_dump_restore_of_simple_db_from_pipe" }
-      let(:fifo_path) { Pathname.new(Dir::Tmpname.create("") {}) }
+    # context "with a pg_dump file from a pipe" do
```
Fryguy, line 25: Do we need this anymore?

spec/postgres_admin_spec.rb
```diff
@@ -58,20 +60,26 @@
         expect(author_count).to eq(2)
         expect(book_count).to   eq(3)
       end
+
+      after do
+        if defined?(PostgresRunner)
+          PostgresRunner.hard_reset
+        else
+          CiPostgresRunner.stop
+          CiPostgresRunner.start
```
Fryguy, line 69: Super-minor, but you could add hard_reset method to CiPostgresRunner for consistency with PostgresRunner.

Fryguy:
Merged...everything else can be follow-up, if needed.

spec/postgres_admin_spec.rb
NickLaMuro: Sure, I can do that in a followup.

spec/postgres_admin_spec.rb
NickLaMuro: Probably not... it was failing, but there is no real reason that it should though.

But it isn't really needed, so it can be tossed.  Will do that in a follow up.

PR #161, opened 2021-04-13
Add database dump/backup/restore to CLI

NickLaMuro:
Built on top of https://github.com/ManageIQ/manageiq-appliance_console/pull/160 and includes CLI commands for the corresponding task done via the interactive console.

### DIFF
--- a/lib/manageiq/appliance_console/cli.rb
+++ b/lib/manageiq/appliance_console/cli.rb
@@ -38,7 +38,23 @@ module ApplianceConsole
     end
 
     def database?
-      options[:standalone] || hostname
+      (options[:standalone] || hostname) && !database_admin?
+    end
+
+    def database_admin?
+      db_dump? || db_backup? || db_restore?
+    end
+
+    def db_dump?
+      options[:dump]
+    end
+
+    def db_backup?
+      options[:backup]
+    end
+
+    def db_restore?
+      options[:restore]
     end
 
     def local_database?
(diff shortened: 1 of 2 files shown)


lib/manageiq/appliance_console/cli.rb
```diff
@@ -344,6 +368,33 @@ def set_replication
...
+      PostgresAdmin.restore(extract_db_opts(options))
+    end
+
+    DB_OPT_KEYS = %i[dbname username password hostname port local_file].freeze
+    def extract_db_opts(options)
+      require 'manageiq/appliance_console/postgres_admin'
+
+      db_opts = {}
+
+      DB_OPT_KEYS.each { |k| db_opts[k] = options[k] if options[k] }
```
Fryguy, line 389: Not sure if slice is available, but is this simpler?  If it's really important that keys without values are removed you could also tack on a `.compact` or `.delete_blanks` or something like that.

```suggestion
      db_opts = options.slice(*DB_OPT_KEYS)
```

Fryguy:
Only one super-minor comment, otherwise LGTM.  Is this ready to go (i.e. can it be un-WIPed)?

Fryguy (approved):

NickLaMuro:
> Is this ready to go (i.e. can it be un-WIPed)?

@Fryguy It is WIP while I get tests in place using this code:

https://github.com/NickLaMuro/manageiq-db_backup-tests/

I have some code locally that is mostly working, but need to push it up, and confirm it works for all cases.

> Only one super-minor comment, otherwise LGTM.

I will take a look at implementing the change and seeing if it is available.  I think we have `ActiveRecord` available though, so my guess would be `ActiveSupport` is there as well.  Will check though.

NickLaMuro:
@Fryguy After some annoying `postgres` connection issues with the secondary DB instance I use for restores, I think I have this working now.

No changes necessary to get the tests passing and using this code, just fixing the test code.  Have it merged over here if you are curious:

https://github.com/NickLaMuro/manageiq-db_backup-tests/pull/1

But I feel more confident merging this now that I have something to validate any future changes with.

PR #163, opened 2021-05-28
[WIP] make v2_key private

kbrock:
not a dependent, but also note we are setting the permissions elsewhere as well as a backup https://github.com/ManageIQ/manageiq-password/pull/22

chmod `v2_key` and `database.yml` so it is accessible by root and manageiq user.
The directory has a `SGID` bit so the files are accessible by manageiq via the group.

~~see also https://github.com/ManageIQ/manageiq/pull/21248~~

### DIFF
--- a/lib/manageiq/appliance_console/database_configuration.rb
+++ b/lib/manageiq/appliance_console/database_configuration.rb
@@ -275,7 +275,7 @@ FRIENDLY
 
     def do_save(settings)
       require 'yaml'
-      File.write(DB_YML, YAML.dump(settings))
+      File.write(DB_YML, YAML.dump(settings), :perm => 0o440)
     end
 
     def initialize_from_hash(hash)
--- a/lib/manageiq/appliance_console/key_configuration.rb
+++ b/lib/manageiq/appliance_console/key_configuration.rb
@@ -73,7 +73,7 @@ module ApplianceConsole
         say("Failed to overwrite original key, original key kept. #{e.message}")
         return false
       end
-      FileUtils.chmod(0o400, KEY_FILE)
+      File.chmod(0o440, KEY_FILE)
     end
 
     def remove_new_key_if_any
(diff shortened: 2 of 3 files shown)


kbrock:
wip: locking down the directory instead

kbrock:
unwip. at least for now, content with the change in chmod and added the database chmod as well.

lib/manageiq/appliance_console/database_configuration.rb
```diff
@@ -275,7 +275,7 @@ def validate_encryption_key!
 
     def do_save(settings)
       require 'yaml'
-      File.write(DB_YML, YAML.dump(settings))
+      File.write(DB_YML, YAML.dump(settings), :perm => 0o440)
```
bdunne, line 278: Should this be 660?

kbrock:
We currently store `v2_key` as `400` and I'm changing that to `440`
Was thinking of doing the same with `database.yml` and the other config files that are more sensitive.

We have locked down `certs/` and `config/` to `770`, so these changes are not strictly necessary. Just feel better with it on the more restrictive side of things.

What ever we decide here, I'll do the same in our rpm. (the upgrade process for rpms ensures that these files are correctly protected)

lib/manageiq/appliance_console/database_configuration.rb
bdunne: After looking into it a bit more, I think 440 is okay, but we do have other writers of this file that probably need to be updated:
- https://github.com/ManageIQ/manageiq/blob/master/lib/manageiq/environment.rb#L43
- https://github.com/ManageIQ/manageiq-postgres_ha_admin/blob/master/lib/manageiq/postgres_ha_admin/config_handler/rails_config_handler.rb#L31
- fix auth

kbrock: feels like we do this stuff in quite a few places.

I was adding the chmod code into our rpm process as well.

v2 key also is handles in fix auth, here, and chmodded in the rpm.
We have other areas that copy the sample files.

Thank you for the links. I'll look into changing all of these or if possible, find a way to reduce all these

kbrock:
Closing

we have locked down the config and certs directory to `manageiq.manageiq 660`
So locking down these individual files is not as necessary as before.
Also, since we are using the user instead of the group for the bulk of our access, changing the privileges from `400` to `440` is no longer necessary

PR #165, opened 2021-07-30
Chown database.yml as manageiq:manageiq

agrare:
Fixes https://github.com/ManageIQ/manageiq/issues/21358

Files which get written as root and cause workers to fail to start up:
- [x] `config/database.yml`
- [x] `config/messaging.yml`
- [x] `certs/v2_key`

### DIFF
--- a/lib/manageiq/appliance_console/message_configuration.rb
+++ b/lib/manageiq/appliance_console/message_configuration.rb
@@ -120,10 +120,8 @@ module ManageIQ
           messaging_yaml["production"]["security.protocol"] = "PLAINTEXT"
         end
 
-        File.open(messaging_yaml_path, "w") do |f|
-          f.write(messaging_yaml.to_yaml)
-          f.chown(manageiq_uid, manageiq_gid)
-        end
+        File.write(messaging_yaml_path, messaging_yaml.to_yaml)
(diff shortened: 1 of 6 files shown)


agrare:
I looks like specs are failing on master but there is no cron job to run travis periodically.

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -117,6 +121,7 @@ def configure_messaging_yaml
         end
 
         File.write(messaging_yaml_path, messaging_yaml.to_yaml)
+        File.chown(manageiq_uid, manageiq_gid, messaging_yaml_path)
```
bdunne, line 124: Should this follow [the same pattern as above](https://github.com/ManageIQ/manageiq-appliance_console/pull/165/files#diff-58f7d485041385eb768bd8d25d7e83a11ccf1a1bbf2bc4fb136fcdcf9cd6b590R282)?

bdunne:
cc @kbrock since we talked about this the other day.

lib/manageiq/appliance_console/message_configuration.rb
agrare: :+1: done

Issue #166, opened 2021-07-30
Failing to configure messaging throws exception


agrare:
```
Configure Messaging Type
/opt/manageiq/manageiq-gemset/gems/manageiq-appliance_console-6.1.0/lib/manageiq/appliance_console/message_configuration.rb:182:in `configure_messaging_type': undefined method `parse_errors' for #<ManageIQ::ApplianceConsole::MessageServerConfiguration:0x000055b81989f630> (NoMethodError)
	from /opt/manageiq/manageiq-gemset/gems/manageiq-appliance_console-6.1.0/lib/manageiq/appliance_console/message_configuration.rb:192:in `unconfigure'
	from /opt/manageiq/manageiq-gemset/gems/manageiq-appliance_console-6.1.0/lib/manageiq/appliance_console/message_configuration_server.rb:101:in `unconfigure'
	from /opt/manageiq/manageiq-gemset/gems/manageiq-appliance_console-6.1.0/lib/manageiq/appliance_console/message_configuration.rb:133:in `valid_environment?'
	from /opt/manageiq/manageiq-gemset/gems/manageiq-appliance_console-6.1.0/lib/manageiq/appliance_console/message_configuration.rb:52:in `ask_questions'
	from /opt/manageiq/manageiq-gemset/gems/manageiq-appliance_console-6.1.0/bin/appliance_console:428:in `block in <module:ApplianceConsole>'
```

PR #169, opened 2021-08-02
Add missing parse_errors() method

agrare:
When configuring the messaging_type fails the `MessageConfiguration` class raises `raise parse_errors(result)` but the parse_errors method didn't exist.

Fixes https://github.com/ManageIQ/manageiq-appliance_console/issues/166

### DIFF
--- a/lib/manageiq/appliance_console/external_auth_options.rb
+++ b/lib/manageiq/appliance_console/external_auth_options.rb
@@ -82,8 +82,7 @@ module ApplianceConsole
         say("\nUpdating external authentication options on appliance ...")
         params = update_hash.collect { |key, value| "#{key}=#{value}" }
         params = configure_provider_type!(params)
-        result = ManageIQ::ApplianceConsole::Utilities.rake_run("evm:settings:set", params)
-        raise parse_errors(result).join(', ') if result.failure?
+        ManageIQ::ApplianceConsole::Utilities.rake_run!("evm:settings:set", params)
       end
     end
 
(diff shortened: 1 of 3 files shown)


agrare:
It looks like this whole thing [[ref]](https://github.com/ManageIQ/manageiq-appliance_console/blob/master/lib/manageiq/appliance_console/message_configuration.rb#L178-L180) was copied from `external_auth_options.rb` [[ref]](https://github.com/ManageIQ/manageiq-appliance_console/blob/master/lib/manageiq/appliance_console/external_auth_options.rb#L85-L87)

We should consider moving parse_errors into that Utilities and/or consider `.run_rake!` which raises the exception automatically

lib/manageiq/appliance_console/utilities.rb
```diff
@@ -16,6 +16,17 @@ def self.rake_run(task, params)
       result
     end
 
+    def self.run_rake!(task, params)
```
chessbyte, line 19: Method name is `run_rake!` but callers use `ManageIQ::ApplianceConsole::Utilities.rake_run!`

lib/manageiq/appliance_console/external_auth_options.rb
```diff
@@ -152,17 +151,8 @@ def self.configured?
...
-
-      if result.success?
-        return parse_response(result)
-      else
-        raise parse_errors(result).join(', ')
-      end
-    end
-
-    def parse_errors(result)
-      result.error.split("\n").collect { |line| line if line =~ /^error: /i }.compact
```
agrare, line 165: Not sure why they were doing `collect` with an `if` and a `.compact` rather than a `select` :shrug:

lib/manageiq/appliance_console/utilities.rb
```diff
@@ -16,6 +16,16 @@ def self.rake_run(task, params)
       result
     end
 
+    def self.rake_run!(task, params)
+      result = run_rake(task, params)
```
chessbyte, line 20: should `rake_run` method be getting called here?

agrare: Yes just a little dyslexic this morning :laughing:

PR #170, opened 2021-08-02
Create /opt/kafka/config/keystore directory

agrare:
Create the keystore directory before writing config files to it.

Fixes https://github.com/ManageIQ/manageiq/issues/21357

### DIFF
--- a/lib/manageiq/appliance_console/message_configuration_client.rb
+++ b/lib/manageiq/appliance_console/message_configuration_client.rb
@@ -82,6 +82,8 @@ module ManageIQ
       def fetch_from_server(src_file, dst_file)
         return if file_found?(dst_file)
 
+        FileUtils.mkdir_p(dst_file.dirname) unless dst_file.dirname.directory?
+
         Net::SCP.start(message_server_host, message_server_username, :password => message_server_password) do |scp|
           scp.download!(src_file, dst_file)
         end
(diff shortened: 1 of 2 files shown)


lib/manageiq/appliance_console/message_configuration_client.rb
```diff
@@ -82,6 +82,8 @@ def fetch_ca_cert_from_server
       def fetch_from_server(src_file, dst_file)
         return if file_found?(dst_file)
 
+        FileUtils.mkdir_p(dst_file.dirname) unless dst_file.dirname.directory?
```
agrare, line 85: This appears to always be a `Pathname` (callers: [`fetch_truststore_from_server`](https://github.com/ManageIQ/manageiq-appliance_console/blob/master/lib/manageiq/appliance_console/message_configuration_client.rb#L71), [`fetch_ca_cert_from_server`](https://github.com/ManageIQ/manageiq-appliance_console/blob/master/lib/manageiq/appliance_console/message_configuration_client.rb#L74-L78))

lib/manageiq/appliance_console/message_configuration_server.rb
```diff
@@ -1,3 +1,4 @@
+
```
chessbyte, line 1: why this newline?

agrare: Leftover from another change I pulled out of this PR

PR #173, opened 2021-09-15
We need to update to at least 6.0.4.1

bdunne:
https://github.com/advisories/GHSA-2rqw-v265-jf8c

### DIFF
--- a/manageiq-appliance_console.gemspec
+++ b/manageiq-appliance_console.gemspec
@@ -20,8 +20,8 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_runtime_dependency "activerecord",            "~> 6.0.3.5"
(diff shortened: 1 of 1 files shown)


manageiq-appliance_console.gemspec
```diff
@@ -20,8 +20,8 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_runtime_dependency "activerecord",            "~> 6.0.3.5"
-  spec.add_runtime_dependency "activesupport",           "~> 6.0.3.5"
+  spec.add_runtime_dependency "activerecord",            "~> 6.0.4"
+  spec.add_runtime_dependency "activesupport",           "~> 6.0.4"
```
Fryguy, line 24: Just to make it clearer, can we do:

`"~> 6.0.4", ">= 6.0.4.1"`

Fryguy: We probably can also do the same on core.

Fryguy (approved):

PR #179, opened 2022-02-18
Disable --message-{server, client}-config

agrare:
Temporarily prevent the ability to configure a kafka message server and client


agrare:
I figured I would leave disable in case someone had already enabled kafka

Fryguy:
I'm not sure how we're going to release this.  Technically this is a breaking change, but maybe we should just leave this and downgrade instead in the application?

Fryguy:
Closing in favor of #180

PR #178, opened 2022-02-15
[WIP] Switch to GitHub Actions

Fryguy:
[WIP] Switch to GitHub Actions

PR #182, opened 2022-07-08
pass region as an environmental variable for db creation

kbrock:
FYI:

- This started https://github.com/ManageIQ/manageiq/pull/6248
- continued through https://github.com/ManageIQ/manageiq/pull/17257
- Where it ends https://github.com/ManageIQ/manageiq/pull/21981

### What

Transitioning to using an environment variable.

### Why

If we use a `REGION` env variable then we can transition from `rake evm:db:region` to  `rake db:migrate` or `rake db:reset`


Fryguy:
@kbrock Can you run specs locally and verify they pass?  I forgot that https://github.com/ManageIQ/manageiq-appliance_console/pull/178 hasn't been merged yet.

Fryguy (approved):
LGTM - not sure if this should be a patch release or a minor release.  I think patch release because we've supported the env var for a LONG time.

kbrock:
tests pass locally.

Does look like we have a macos only issue since `df -T -P` isn't quite the same thing on bsd and linux. But that is an issue not introduced nor fixed in this PR

PR #187, opened 2022-08-12
Introduce CI via actions

kbrock:


kbrock:
ok. looks like we already have this in progress

PR #189, opened 2022-09-01
6.1 configurations has symbol keys, [] deprecated, use configs_for

jrafanie:
Fixes https://github.com/ManageIQ/manageiq-appliance/issues/366

DEPRECATION WARNING: [] is deprecated and will be removed from Rails 6.2 (Use configs_for)

This fixes the deprecation
This now uses symbols for configuration keys (since rails 6.1 has changed this as well)

Similar to: https://github.com/ManageIQ/manageiq/pull/21652/commits/0d57f5efbba53b5b27656f81922031320691750a

### DIFF
--- a/lib/manageiq/appliance_console/utilities.rb
+++ b/lib/manageiq/appliance_console/utilities.rb
@@ -27,8 +27,13 @@ module ApplianceConsole
     end
 
     def self.db_connections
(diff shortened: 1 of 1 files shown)


jrafanie:
cc @kbrock

jrafanie:
I'm guessing we'll need a new release of this so we can use it going forward

PR #190, opened 2022-09-09
Report messaging configuration for application configuration

nasark:
- require users to configure messaging if they do not already have it configured when setting up the application

@miq-bot assign @agrare

### DIFF
--- a/bin/appliance_console
+++ b/bin/appliance_console
@@ -92,18 +92,20 @@ module ApplianceConsole
       eth0.reload
       eth0.parse_conf if eth0.respond_to?(:parse_conf)
 
-      host        = LinuxAdmin::Hosts.new.hostname
-      ip          = eth0.address
-      mac         = eth0.mac_address
-      mask        = eth0.netmask
-      gw          = eth0.gateway
-      dns1, dns2  = dns.nameservers
-      order       = dns.search_order.join(' ')
-      timezone    = LinuxAdmin::TimeDate.system_timezone
-      version     = File.read(VERSION_FILE).chomp if File.exist?(VERSION_FILE)
-      dbhost      = ManageIQ::ApplianceConsole::DatabaseConfiguration.database_host
-      database    = ManageIQ::ApplianceConsole::DatabaseConfiguration.database_name
-      evm_running = LinuxAdmin::Service.new("evmserverd").running?
+      host             = LinuxAdmin::Hosts.new.hostname
+      ip               = eth0.address
+      mac              = eth0.mac_address
+      mask             = eth0.netmask
+      gw               = eth0.gateway
+      dns1, dns2       = dns.nameservers
+      order            = dns.search_order.join(' ')
+      timezone         = LinuxAdmin::TimeDate.system_timezone
+      version          = File.read(VERSION_FILE).chomp if File.exist?(VERSION_FILE)
+      dbhost           = ManageIQ::ApplianceConsole::DatabaseConfiguration.database_host
+      database         = ManageIQ::ApplianceConsole::DatabaseConfiguration.database_name
+      evm_running      = LinuxAdmin::Service.new("evmserverd").running?
+      messaging        = ManageIQ::ApplianceConsole::MessageConfiguration.configured?
+      messaging_broker = ManageIQ::ApplianceConsole::MessageServerConfiguration.configured?
 
       summary_attributes = [
         summary_entry("Hostname", host),
@@ -120,6 +122,8 @@ module ApplianceConsole
         summary_entry("#{I18n.t("product.name")} Server", evm_running ? "running" : "not running"),
         summary_entry("#{I18n.t("product.name")} Database", dbhost || "not configured"),
         summary_entry("Database/Region", database ? "#{database} / #{region.to_i}" : "not configured"),
+        summary_entry("Messaging", messaging ? "configured" : "not configured"),
+        summary_entry("Local Messaging Broker", messaging_broker ? "configured" : "not configured"),
         summary_entry("External Auth", ExternalHttpdAuthentication.config_status),
         summary_entry("#{I18n.t("product.name")} Version", version),
       ]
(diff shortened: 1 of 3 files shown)


bin/appliance_console
```diff
@@ -454,6 +455,12 @@ Static Network Configuration
             press_any_key
             raise MiqSignalError
           end
+        else
+          if !MessageConfiguration.configured?
+            say("\nApplication configuration failed, Messaging configuration is required\n")
+            press_any_key
+            raise MiqSignalError
+          end
```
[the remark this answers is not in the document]
nasark: It will throw the error if the user selects `make no messaging changes` while messaging has not already been configured `!MessageConfiguration.configured?`

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -26,6 +26,10 @@ def self.available?
         File.exist?("#{BASE_DIR}/bin/kafka-run-class.sh")
       end
 
+      def self.configured?
+        MessageServerConfiguration.configured? || MessageClientConfiguration.configured?
+      end
+
```
nasark, line 32: Would it be better to use `&&` here so that we return messaging as configured if both the client and server are configured?

agrare: There should only be one appliance in a region configured as a server and the rest should be clients so `||` is fine.

bin/appliance_console
```diff
@@ -117,6 +118,7 @@ module ApplianceConsole
         summary_entry("MAC Address", mac),
         summary_entry("Timezone", timezone),
         summary_entry("Local Database Server", PostgresAdmin.local_server_status),
+        summary_entry("Messaging", messaging ? "configured" : "not configured"),
```
Fryguy, line 121: This ends up in the middle of the database information (the next 3 lines are DB related), so I think we should move it after that (before `"External Auth"`)

bin/appliance_console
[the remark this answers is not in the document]
agrare: Sounds good, I wonder if we should do this in a few steps though.  First just report if messaging is configured and I think we can merge that at any time.

Starting to require this to be set up is going to have to be merged in sync with a number of other items.

What do you think about dropping this block and simply reporting if messaging is configured or not in appliance console?  Maybe have a "local messaging broker" line similar to the local database one that checks if it is a server specifically.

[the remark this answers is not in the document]
nasark: Sounds like a good idea, I'll make the change

nasark:
@agrare I've kept `MessageConfiguration.configured?` and `MessageClientConfiguration.configured?` as helper methods that could be used in the future, but let me know if they should be removed since they are not being used

bin/appliance_console
```diff
@@ -120,6 +121,7 @@ module ApplianceConsole
         summary_entry("#{I18n.t("product.name")} Server", evm_running ? "running" : "not running"),
         summary_entry("#{I18n.t("product.name")} Database", dbhost || "not configured"),
         summary_entry("Database/Region", database ? "#{database} / #{region.to_i}" : "not configured"),
+        summary_entry("Local Messaging Broker", messaging ? "configured" : "not configured"),
```
agrare, line 124: This looks good, think it would be helpful to know if messaging is configured in general in addition to if the broker is running locally:
```suggestion
        summary_entry("Messaging", ManageIQ::ApplianceConsole::MessageConfiguration.configured? ? "configured" : "not configured")
        summary_entry("Local Messaging Broker", messaging ? "configured" : "not configured"),
```

bin/appliance_console
```diff
@@ -104,6 +104,7 @@ module ApplianceConsole
       dbhost      = ManageIQ::ApplianceConsole::DatabaseConfiguration.database_host
       database    = ManageIQ::ApplianceConsole::DatabaseConfiguration.database_name
       evm_running = LinuxAdmin::Service.new("evmserverd").running?
+      messaging   = ManageIQ::ApplianceConsole::MessageServerConfiguration.configured?
```
Fryguy, line 107: Is this used or was this supposed to be used on line 124 below?

nasark: I've updated var names for clarity. `messaging` -> `messaging_broker`

bin/appliance_console
```diff
@@ -92,18 +92,20 @@ module ApplianceConsole
...
+      mask             = eth0.netmask
+      gw               = eth0.gateway
+      dns1, dns2       = dns.nameservers
+      order            = dns.search_order.join(' ')
+      timezone         = LinuxAdmin::TimeDate.system_timezone
+      version          = File.read(VERSION_FILE).chomp if File.exist?(VERSION_FILE)
+      dbhost           = ManageIQ::ApplianceConsole::DatabaseConfiguration.database_host
+      database         = ManageIQ::ApplianceConsole::DatabaseConfiguration.database_name
+      evm_running      = LinuxAdmin::Service.new("evmserverd").running?
+      messaging        = ManageIQ::ApplianceConsole::MessageConfiguration.configured
```
Fryguy, line 107:
```suggestion
      messaging        = ManageIQ::ApplianceConsole::MessageConfiguration.configured?
```

nasark: Thanks!

Fryguy:
Can you run the appliance console specs locally and verify they pass (GHA doesn't work on this repo yet)

agrare:
> Can you run the appliance console specs locally and verify they pass (GHA doesn't work on this repo yet)

:scream: well specs on master fail locally, merging this PR locally doesn't introduce any _new_ failures

Fryguy:
Yeah fair point - I shoudl have mentioned that :)

agrare:
Okay I was trying to get the tests green first then merge this but that's going to be more work than I thought haha

On master:
```
Finished in 1.87 seconds (files took 0.97014 seconds to load)
531 examples, 48 failures
```

With this merged:
```
Finished in 1.54 seconds (files took 0.7703 seconds to load)
531 examples, 48 failures
```

PR #192, opened 2022-09-28
Make backward compatible changes to work with repmgr13 - version 5.2.1

jrafanie:
We're upgrading from 4.0.6 to 5.2.1 along with PG 10 to PG 13.

See: ManageIQ/manageiq-rpm_build#309

* configuration: single quote strings... don't leave them unquoted to be safe
* recovery.conf is gone in PG 12+ / use standby.signal instead to detect a standby (or lack of it for a primary)
* Keep compatibility with repmgr10 for now

### DIFF
--- a/lib/manageiq/appliance_console/postgres_admin.rb
+++ b/lib/manageiq/appliance_console/postgres_admin.rb
@@ -62,13 +62,18 @@ module ApplianceConsole
       LinuxAdmin::Service.new(service_name).running?
     end
 
-    def self.local_server_in_recovery?
-      data_directory.join("recovery.conf").exist?
+    def self.local_server_received_standby_signal?
+      # Beginning with PostgreSQL 12, replication configuration has been integrated into the main PostgreSQL configuraton system and the conventional recovery.conf file is no longer valid.
+      # see: https://repmgr.org/docs/current/release-5.0.html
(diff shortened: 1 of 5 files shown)


lib/manageiq/appliance_console/postgres_admin.rb
```diff
@@ -62,13 +62,18 @@ def self.service_running?
       LinuxAdmin::Service.new(service_name).running?
     end
 
-    def self.local_server_in_recovery?
-      data_directory.join("recovery.conf").exist?
+    def self.local_server_received_standby_signal?
+      # Beginning with PostgreSQL 12, replication configuration has been integrated into the main PostgreSQL configuraton system and the conventional recovery.conf file is no longer valid.
+      # see: https://repmgr.org/docs/current/release-5.0.html
+      # https://www.2ndquadrant.com/en/blog/replication-configuration-changes-in-postgresql-12/
+      # “standby.signal” – indicates the server should start up as a hot standby
```
Fryguy, line 69: Prefer "normal" double quotes so we don't have special characters in the source.

```suggestion
      # "standby.signal" – indicates the server should start up as a hot standby
```

jrafanie:
Ok, I've retested this pull request and verified that it works with repmgr10 on our existing centos 8 / pg 10 appliances.

PR #194, opened 2022-10-11
Extract handling of evmserverd actions to a class

agrare:
Working with starting, stopping, restarting, etc... the evmserverd service was scattered throughout the appliance_console code.  Extract all of this to a service so that we can centralize logic.

PR #195, opened 2022-10-11
[WIP] Only start evmserverd after db and messaging config

agrare:
Depends on:
- [ ] https://github.com/ManageIQ/manageiq-appliance_console/pull/194


Fryguy:
Merged #194 , so this can be rebased.

bin/appliance_console
```diff
@@ -393,6 +393,14 @@ Static Network Configuration
 
         messaging_action = ask_with_menu("Configure Messaging", messaging_options)
 
+        no_changes = database_action == "no_changes" && messaging_action == "no_changes"
+
+        # Stop evmserver while we make changes to the database and/or messaging configuration
+        unless no_changes
```
Fryguy, line 399: The double negative is killing me 😆

bin/appliance_console
```diff
@@ -393,6 +393,14 @@ Static Network Configuration
 
         messaging_action = ask_with_menu("Configure Messaging", messaging_options)
 
+        no_changes = database_action == "no_changes" && messaging_action == "no_changes"
```
Fryguy, line 396: Perhaps, instead we make it a positive variable?

```suggestion
        changes_requested = database_action != "no_changes" || messaging_action != "no_changes"
```

Fryguy: Or

```suggestion
        changes_requested = !(database_action == "no_changes" && messaging_action == "no_changes")
```

agrare: :+1: done

PR #196, opened 2022-10-13
Don't start evmserverd if messaging is not configured

agrare:
Only start evmserverd once database and messaging have been configured

NOTE added a third state to `ManageIQ Server`: `running, not running, not configured` so it is consistent with the database status

image

image

Depends on:
- [x] https://github.com/ManageIQ/manageiq-appliance_console/pull/195

Required for:
* https://github.com/ManageIQ/manageiq/issues/22132

### DIFF
--- a/lib/manageiq/appliance_console/evm_server.rb
+++ b/lib/manageiq/appliance_console/evm_server.rb
@@ -1,21 +1,30 @@
 module ManageIQ
   module ApplianceConsole
     class EvmServer
+      class NotRunnableError < RuntimeError; end
+
       class << self
         def running?
           service.running?
         end
 
         def start!(enable: false)
+          raise NotRunnableError, "Cannot start #{I18n.t("product.name")}: #{not_runnable_reason}" unless runnable?
+
           service.start(enable)
         end
 
         def start(enable: false)
           start!(:enable => enable)
         rescue AwesomeSpawn::CommandResultError => e
-          say e.result.output
-          say e.result.error
-          say ""
+          say(e.result.output)
+          say(e.result.error)
+          say("")
+          false
+        rescue NotRunnableError => e
+          say(e.to_s)
+          say("")
+          false
         end
 
         def stop
@@ -32,6 +41,18 @@ module ManageIQ
 
         def disable
           service.disable
+        end
+
+        def runnable?
+          DatabaseConfiguration.database_yml_configured? && MessageConfiguration.configured?
+        end
+
+        def not_runnable_reason
+          if !DatabaseConfiguration.database_yml_configured?
+            "A Database connection has not been configured."
+          elsif !MessageConfiguration.configured?
+            "Messaging has not been configured."
+          end
         end
 
         private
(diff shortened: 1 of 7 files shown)


lib/manageiq/appliance_console/evm_server.rb
```diff
@@ -0,0 +1,65 @@
...
+
+          service.start(enable)
+        end
+
+        def start(enable: false)
+          start!(:enable => enable)
+        rescue AwesomeSpawn::CommandResultError, NotRunnableError => e
+          say e.result.output
+          say e.result.error
+          say ""
```
Fryguy, line 22: where is `say` defined? or is it a global method?

Fryguy:
Love the refactoring, but wondering if that should be a separate PR?  I'm having trouble separating the refactor from the "Dont start evmserverd if messaging not configured"

Fryguy:
Oh nvm, I didn't see that the "depends on" actually was "built on" (I assumed it was an independent PR).

agrare:
@Fryguy you mean like https://github.com/ManageIQ/manageiq-appliance_console/pull/194 ?

lib/manageiq/appliance_console/evm_server.rb
agrare: I see this used everywhere, can check where it is defined though

Fryguy: I assume it comes from highline globally.

Fryguy:
#195 is merged

lib/manageiq/appliance_console/evm_server.rb
```diff
@@ -1,21 +1,30 @@
...
           service.start(enable)
         end
 
         def start(enable: false)
           start!(:enable => enable)
         rescue AwesomeSpawn::CommandResultError => e
-          say e.result.output
-          say e.result.error
-          say ""
+          say(e.result.output)
```
Fryguy, line 20: Were the parens added here for rubocop?  If so, I'd also be ok with adding these to the exclude list in manageiq-style: https://github.com/ManageIQ/manageiq-style/blob/master/styles/base.yml#L123 (perhaps in a new section for highligh / appliance console)

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -187,12 +187,6 @@ def host_reachable?(host, what)
         true
       end
 
-      def configure_messaging_type(value)
-        say(__method__.to_s.tr("_", " ").titleize)
-
-        ManageIQ::ApplianceConsole::Utilities.rake_run!("evm:settings:set", ["/prototype/messaging_type=#{value}"])
-      end
-
```
agrare, line 195: NOTE this will be the default (once https://github.com/ManageIQ/manageiq/pull/21997 is merged) and will not need to be set by appliance console

lib/manageiq/appliance_console/evm_server.rb
agrare: Yes and no, we actually use `say("")` more often than we use `say ""`
```
$ grep -r --exclude-dir=spec/ 'say(' |wc -l
252
$ grep -r --exclude-dir=spec/ 'say ' | wc -l
32
```

So I was trying to be more consistent with the more common calling convention

Fryguy: Sounds good :+1:

PR #197, opened 2022-10-17
Simplify messaging options by saving in yml files

agrare:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/22171

### DIFF
--- a/lib/manageiq/appliance_console/message_configuration.rb
+++ b/lib/manageiq/appliance_console/message_configuration.rb
@@ -112,14 +112,11 @@ module ManageIQ
 
         messaging_yaml = YAML.load_file(messaging_yaml_sample_path)
 
-        messaging_yaml["production"].delete("username")
-        messaging_yaml["production"].delete("password")
-
-        messaging_yaml["production"]["hostname"]          = message_server_host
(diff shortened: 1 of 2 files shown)


lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -112,14 +112,11 @@ def configure_messaging_yaml
 
         messaging_yaml = YAML.load_file(messaging_yaml_sample_path)
 
-        messaging_yaml["production"].delete("username")
-        messaging_yaml["production"].delete("password")
```
agrare, line 116: These two aren't used at all and don't need to be in the template yaml file

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -112,14 +112,11 @@ def configure_messaging_yaml
...
-
-        messaging_yaml["production"]["hostname"]          = message_server_host
```
agrare, line 118: We were setting the `hostname` key then later in MiqQueue changing it to `host` here https://github.com/ManageIQ/manageiq/pull/22171/files#diff-0e5efff55cd8273572873c343f228135e77fd302231c419486fa601e343af459L693

PR #198, opened 2022-10-17
Swap "localhost" to 127.0.0.1 in Messaging server configuration

nasark:
- resolves https://github.com/ManageIQ/manageiq/issues/22153

```
[root@9-46 ~]# systemctl status kafka
● kafka.service - kafka service
   Loaded: loaded (/usr/lib/systemd/system/kafka.service; enabled; vendor preset: disabled)
   Active: active (running) since Mon 2022-10-17 14:33:38 EDT; 2min 59s ago
 Main PID: 98945 (java)
    Tasks: 76 (limit: 101046)
   Memory: 410.0M
   CGroup: /system.slice/kafka.service
           └─98945 java -Xmx1G -Xms1G -server -XX:+UseG1GC -XX:MaxGCPauseMillis=20 -XX:InitiatingHeapOccupancyPercent=35 -XX:+ExplicitGCInvokesConcurrent -XX:MaxInlineLevel=15 -Djava.awt.headless=true -Xloggc:/var/log/kafka/kafkaServer-gc.log -verbose:gc -XX:+PrintGCDeta>

Oct 17 14:33:42 9-46.68.56.rtp.raleigh.ibm.com kafka-server-start.sh[98945]: [2022-10-17 14:33:42,460] INFO [GroupMetadataManager brokerId=0] Finished loading offsets and group metadata from __consumer_offsets-12 in 55 milliseconds for epoch 0, of which 55 milliseconds w>
Oct 17 14:33:42 9-46.68.56.rtp.raleigh.ibm.com kafka-server-start.sh[98945]: [2022-10-17 14:33:42,460] INFO [GroupMetadataManager brokerId=0] Finished loading offsets and group metadata from __consumer_offsets-21 in 54 milliseconds for epoch 0, of which 54 milliseconds w>
Oct 17 14:33:42 9-46.68.56.rtp.raleigh.ibm.com kafka-server-start.sh[98945]: [2022-10-17 14:33:42,460] INFO [GroupMetadataManager brokerId=0] Finished loading offsets and group metadata from __consumer_offsets-36 in 54 milliseconds for epoch 0, of which 54 milliseconds w>
Oct 17 14:33:42 9-46.68.56.rtp.raleigh.ibm.com kafka-server-start.sh[98945]: [2022-10-17 14:33:42,460] INFO [GroupMetadataManager brokerId=0] Finished loading offsets and group metadata from __consumer_offsets-6 in 54 milliseconds for epoch 0, of which 54 milliseconds wa>
[...]

PR #199, opened 2022-10-18
Release v8.0.0

agrare:
I think we should release an 8.0 so the messaging changes aren't pulled into any current release branches (oparin is on `"~>7.1", ">=7.1.1"` right now).

We didn't have a CHANGELOG here so I added one with the 8.0 changes we can back fill it if we want.

TODO before release:
- [x] https://github.com/ManageIQ/manageiq-appliance_console/pull/197
- [x] https://github.com/ManageIQ/manageiq-appliance_console/pull/198

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,21 @@
+# Change Log
+All notable changes to this project will be documented in this file.
+This project adheres to [Semantic Versioning](http://semver.org/).
+
+## [Unreleased]
+
+## [8.0.0] - 2022-10-18
+### Fixed
+- Don't require pressing any key twice for message configuration (#193)
+
+### Added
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
+# Change Log
+All notable changes to this project will be documented in this file.
+This project adheres to [Semantic Versioning](http://semver.org/).
+
+## [Unreleased]
+
+## [8.0.0] - 2022-10-18
```
Fryguy, line 7: It's unclear here why we are cutting an 8.0 - wondering if we should add some **BREAKING** lines about the messaging stuff?  I think technically we are making breaking changes by enforcing messaging.

agrare:
Rebased and taken out of WIP

PR #200, opened 2023-01-03
Upgrade to highline 2.1.0

bdunne:
https://github.com/JEG2/highline/pull/255/files#diff-c275fec9c453c2a42515bc5ab47e30fa4130ff1426aef2f6e000a9a34e122cb8R9

Includes https://github.com/JEG2/highline/pull/250

PR #201, opened 2023-01-04
Update to Highline 2.1.0

### LINKED ISSUE #181: Upgrade highline gem
highline 1.6.21 has an LGPL file shipped within, which creates a hassle for license scanning.  The newer versions do not have this file, so let's see if we can upgrade it.

kbrock:
steal anything you want out of here @bdunne 

not thrilled with the changes to tests with the expect heard stuff.
wish the questions were not going through, but this is the least invasive way I could think to solve

overlaps with #200 

Fixes #181


spec/database_admin_spec.rb
```diff
@@ -102,7 +102,7 @@
           expect(subject.ask_file_location).to be_truthy
 
           error = "Please provide #{errmsg}"
-          expect_heard ["Enter the #{prmpt}: ", error, prompt]
+          expect_heard ["Enter the #{prmpt}: |#{"/tmp/evm_db.backup"}| ", error, prompt]
```
Fryguy, line 105:
```suggestion
          expect_heard ["Enter the #{prmpt}: |/tmp/evm_db.backup| ", error, prompt]
```

kbrock:
update:
- fixed highline gem value
- removed string interpolation in test
- moved dependency to spot where dependencies were defined (it was not finding the files for some reason, but it may have been another problem. seemed good to me anyway)

I tested on a nightly appliance. had to muck with the appliance console gemspec in 2 places and the vmdb/gemfile.lock, but I was able to setup dhcp, create a v2 key, set the region and configure an internal database.

kbrock:
@bdunne let me know if you are happy with these changes.
as I mentioned I was able to change dhcp, and setup a basic appliance. If there are other configuration options you want me to test, let me know.

kbrock:
@Fryguy or @bdunne I did run through those few tests. Anything else we want to do to be comfortable merging?

Fryguy (approved):
LGTM - @bdunne Please also review.

Fryguy:
@kbrock Can you run the full rspec suite locally?  We still don't have github actions working in this repository yet.

Fryguy:
Bouncing now that GitHub Actions is in place - 3.0 should still fail cause of kwargs at the moment

kbrock:
update:
- removed the pending
update:
- removed the CI tty hack (still needed)
update
- testing out alternative no-echo input

Fryguy (approved):
LGTM

Fryguy:
Love that you could kill the tty hack too

Issue #205, opened 2023-03-21
Post GitHub Actions cleanup


Fryguy:
- [x] Fix tests on Ruby 3.0 and remove fail-fast #206
- [ ] Undo pending specs
  - [ ] PostgresRunner
  - [x] spec/database_admin_spec.rb fails for some weird stream issue only on CI and in act with `true | act 2>&1 | cat`  Removed in #201
- [x] Perhaps remove the TTY hacks?  This might be easier or not needed after #201

kbrock:
may want to consider #207

PR #206, opened 2023-03-21
support ruby 3.0

kbrock:
part of #205 

rspec expectations.with now needs hashes to be explicitly passed as a hash


Fryguy (approved):

kbrock:
update:
- removed the fail-fast: false

PR #207, opened 2023-03-21
[WIP] Avoid Readline when in non-tty

kbrock:
We have 3 prompt methods. Change all to be more tty friendly:

- any_key: use input and not STDIN
- press_ask_yn: don't use readline (we were sporadic on this support anyway)
- just_ask: only use readline if on a tty.

for ask_yn, we could convert over to all using readln, but this required test changes

### DIFF
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -127,15 +127,11 @@ module ManageIQ
       end
 
       def should_exclude_tables?
-        ask_yn?("Would you like to exclude tables in the dump") do |q|
-          q.readline = true
-        end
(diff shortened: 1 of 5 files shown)


Fryguy:
@kbrock Have you tested the appliance console manually with these changes on an appliance?

kbrock:
WIP: need to manually test
(we need to test the other changes too, so all together sounds like a good plan)

kbrock:
Punting on this for now.

They released highline 3.0.0.pre1
This incorporated my tty changes and helps us avoid hacking it on our side.

I'm leaving this PR open only so we have a reminder that we need to upgrade (and test) the new highline.

kbrock:
going with #210 instead

PR #208, opened 2023-04-21
Move gem name loader to proper namespaced location.

jrafanie:
Move gem name loader to proper namespaced location.


kbrock:
I think the proper solution is to create `lib/manageiq-appliance_console.rb` that requires `manageiq/appliance_console`

This is the pattern I tend to see in gems.

jrafanie:
@kbrock yeah, I understand that is/was the standard for non-rails stuff but zeitwerk is adopting the fairly new (maybe last several years) standard of lib/manageiq/appliance_console.rb, see: https://guides.rubygems.org/name-your-gem/

Checkout https://github.com/fxn/zeitwerk/issues 138 for more background.

Basically, zeitwerk is expecting files to define the constants based on their name and these "gem name.rb" loader files in lib fail this test.  I'd like to find a way to get zeitwerk to work with them without needing to continually tell zeitwerk to ignore them but for now, this is the easiest way.

Basically, unless we expect requires to use the manageiq-appliance_console format, we can safely let bundler do the require as it will try the hyphen form, and if that doesn't exsit, it will use the manageiq/appliance_console form of require.

kbrock:
searched all the code. Looks like the only references to `require "manageiq-appliance_console"` are from within this gem.

kbrock:
kicking to start the build again.

kbrock:
@jrafanie I added #209 to fix the build.
After we merge that, lets see if this turns green.

PR #209, opened 2023-04-27
Remove miq specific source

kbrock:
Fixes broken CI failure

We need to have only one main source in a `Gemfile`

```
[DEPRECATED] Your Gemfile contains multiple primary sources.
Using `source` more than once without a block is a security risk, and may result in installing unexpected gems.
To resolve this warning, use a block to indicate which gems should come from the secondary source.
```

- We added this source for `handsoap` dependency in #109 to fix a build failure.
- I'm not able to find when we removed this dependency, and removing it no longer breaks the build.
- Also our `Gemfile.lock` does not mention any manageiq dependencies (besides `manageiq-password`) so I don't see a way we could inadvertently require `handsoap`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,3 @@
-source 'http://rubygems.manageiq.org'
 source 'https://rubygems.org'
 
 # Specify your gem's dependencies in manageiq-appliance_console.gemspec


Fryguy:
LGTM, but do you know why was it added in the first place?

kbrock:
@Fryguy thanks for keeping me honest. I added the notes

jrafanie:
> * I'm not able to find when we removed this dependency, and removing it no longer breaks the build.
> * Also our `Gemfile.lock` does not mention any manageiq dependencies (besides `manageiq-password`) so I don't see a way we could inadvertently require `handsoap`.

Thanks, I found it after you did most of the work.

At the time, this repo had manageiq-gems-pending in the Gemfile:
https://github.com/ManageIQ/manageiq-appliance_console/pull/109/files

image

This gem included handsoap until the vmware stuff was extracted: https://github.com/ManageIQ/manageiq-gems-pending/pull/169

Also, we dropped the gems pending dependency from the gemfile in this repo here: https://github.com/ManageIQ/manageiq-appliance_console/pull/153

jrafanie (approved):
LGTM, thanks for digging in and figuring it out

PR #210, opened 2023-04-28
Clean up test output (highline and stdout messages)

kbrock:
DEPENDENCY:

- [x] first commit (for 🍏 build) #209

This PR:

- Remove messages printed to stdout in the test run.
- Updating our use of `STDIN` to `input` to be more highline 3.0 (next version) friendly.

### DIFF
--- a/spec/support/ui.rb
+++ b/spec/support/ui.rb
@@ -29,10 +29,9 @@ shared_context 'with a terminal ui', :with_ui do
   after do
     @temp_stdin.close! if @temp_stdin
     @temp_stdout.close! if @temp_stdout
-    HighLine.default_instance = HighLine.new(STDIN, STDOUT)
-    # best-guess cleanup: Readline has .input=, .output= but no .input, .output
-    Readline.input = STDIN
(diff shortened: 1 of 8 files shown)


spec/support/ui.rb
```diff
@@ -21,13 +21,15 @@
...
   before do
+    HighLine.default_instance = HighLine.new(input, output)
     Readline.input = input
     Readline.output = readline_output
   end
 
   after do
     @temp_stdin.close! if @temp_stdin
     @temp_stdout.close! if @temp_stdout
+    HighLine.default_instance = HighLine.new(STDIN, STDOUT)
```
kbrock, line 32: not sure exactly what we want here, I was just mimicking the `ReadLine` reset down below.

Fryguy: Perhaps just this?

https://github.com/JEG2/highline/blob/v2.1.0/lib/highline.rb#L648

kbrock:
Update:
- Changed reset of highline to just use `Highline.new`
- Changed calls to reset `Readline#input`

PR #215, opened 2023-06-22
Separate kerberos from service principal name and use correctly

jrafanie:
Fix incomplete fix in #211.

We have two concepts that were being shared.
* kerberos principal name
* service principal name

getcert requires the kerberos principal name with the kerberos realm included: 

`getcert request -K SERVICE/host@REALM`

See https://github.com/ManageIQ/manageiq-appliance_console/blob/9ce14c3087930322bbeac0e2f5a9723d92eea71a/lib/manageiq/appliance_console/certificate.rb#L143-L149 for usage.

ipa service-find and service-add use the service principal name, which doesn't include the kerberos realm as that's assumed based on configuration and cannot be changed without changing the configuration:

`ipa service-find  --principal SERVICE/host`

This commit clarifies these differences and uses the correct mechanism for service-add, service-find, and getcert.

from: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_identity_management/managing-hosts-cli_configuring-and-managing-idm
Monosnap Chapter 42  Managing Hosts in IdM CLI Red Hat Enterprise Linux 8 | Red Hat Customer Portal 2023-06-22 11-07-03

### DIFF
--- a/lib/manageiq/appliance_console/principal.rb
+++ b/lib/manageiq/appliance_console/principal.rb
@@ -10,12 +10,14 @@ module ApplianceConsole
     attr_accessor :service
     # kerberos principal name
     attr_accessor :name
+    attr_accessor :service_principal
 
     def initialize(options = {})
       options.each { |n, v| public_send("#{n}=", v) }
       @ca_name ||= "ipa"
       @realm = @realm.upcase if @realm
-      @name ||= "#{service}/#{hostname}\\@#{realm}"
+      @service_principal ||= "#{service}/#{hostname}"
+      @name ||= "#{service_principal}@#{realm}"
     end
 
     def register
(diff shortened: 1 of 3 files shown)


kbrock (approved):
LGTM. just from a code perspective and a general overview.
I like how both of the specs are more similar to how they were before 211 - with the unescaped `@`.

More than happy to merge. Not sure if you need something other than a pure code review. (I don't know how to setup ipa)

jrafanie:
> LGTM. just from a code perspective and a general overview.
> I like how both of the specs are more similar to how they were before 211 - with the unescaped `@`.

yeah, I agree.  #211 assumed the usage with getcert but not ipa service-add|find.  I believe at some point service add/find no longer accepted the realm in the command line argument.  At that point, we couldn't share the same value with the two consumer client calls.

PR #220, opened 2023-09-22
Add manageiq user to allowed_uids for sssd

jrafanie:
When we moved to using a manageiq user, we need to add this user so it has permission in sssd.conf.

See also https://github.com/ManageIQ/manageiq-documentation/pull/1743

### DIFF
--- a/lib/manageiq/appliance_console/external_httpd_authentication/external_httpd_configuration.rb
+++ b/lib/manageiq/appliance_console/external_httpd_authentication/external_httpd_configuration.rb
@@ -124,7 +124,7 @@ module ApplianceConsole
       def configure_sssd_ifp(config)
         user_attributes = LDAP_ATTRS.keys.collect { |k| "+#{k}" }.join(", ")
         ifp_config      = "
-  allowed_uids = #{APACHE_USER}, root
+  allowed_uids = #{APACHE_USER}, root, manageiq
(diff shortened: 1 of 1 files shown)


Fryguy:
I think there should also be a corresponding change in the ManageIQ operator, since we have these configs there as well.

jrafanie:
> I think there should also be a corresponding change in the ManageIQ operator, since we have these configs there as well.

I think that's here:  https://github.com/ManageIQ/httpd_configmap_generator/pull/77. Right?

Do you know how it's pulled in?  I assume we'll need a new tag in httpd_configmap_generator and pull that in.

jrafanie:
> I think that's here: [ManageIQ/httpd_configmap_generator#77](https://github.com/ManageIQ/httpd_configmap_generator/pull/77). Right?
> 
> Do you know how it's pulled in? I assume we'll need a new tag in httpd_configmap_generator and pull that in.

bump... @bdunne maybe you know?  Do we need to do something to pull in the new configmap_generator?  If we want to backport this, do we need a new version there and here?

Fryguy:
I think I was thinking of https://github.com/ManageIQ/manageiq-pods/blob/master/manageiq-operator/api/v1alpha1/helpers/miq-components/httpd_conf.go ?  The httpd_configmap_generator change I think covers what I was thinking.

PR #221, opened 2023-10-20
Both pg_dump and pg_basebackup are supported ways to backup the database

jrafanie:
There are pros and cons to both logical and physical backups and in fact people should probably plan on doing both types of backups.

Physical/pg_dump for table or database exports for testing, upgrades, and for disaster recovery. It's slower but generally smaller.

Logical/pg_basebackup for cluster backups and versatility to restore from point in time or from standbys. It's also much faster.

See also:
https://www.enterprisedb.com/postgresql-database-backup-recovery-what-works-wal-pitr

### DIFF
--- a/lib/manageiq/appliance_console/database_admin.rb
+++ b/lib/manageiq/appliance_console/database_admin.rb
@@ -15,14 +15,6 @@ module ManageIQ
         Example: 'mydomain.com/user'
       PROMPT
 
-      DB_DUMP_WARNING = <<-WARN.strip_heredoc
-        WARNING:  This is not the recommended and supported way of running a
-        database backup, and is strictly meant for exporting a database for
-        support/debugging purposes!
-
-
-      WARN
-
       attr_reader :action, :backup_type, :database_opts, :delete_agree, :filename
 
       def initialize(action = :restore, input = $stdin, output = $stdout)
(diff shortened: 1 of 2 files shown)


kbrock (approved):
It looks like this may have been related to when we had multiple database for embedded ansible awx. That is no longer valid.

The admins will need to perform this when upgrading postgres versions. So discouraging them from doing a necessary operation is unwise.

Since this PR only removes an incorrect warning, I vote yes.

bdunne:
Could this have anything to do with HA?

jrafanie:
> Could this have anything to do with HA?

@bdunne Yes.  I think it's more a question of how granular do you want your DR solution to be.  Basebackup will capture all of the HA configuration along with region replication configuration and our data but also offer point in time recovery so it's far more extensive in terms of what it offers.  But, it assumes you can rebuild the environment exactly as before with the same hostnames/ipaddresses and that you'll be using the basebackup backup on the same version.  So, if you want a full DR solution, it's superior, but it's also more complicated because you have to be familiar with using basebackup and recovering to the specific time you want.  I think that's outside the scope of what we can advise them on.

pg_dump is pretty simplistic in that it's not a complete backup but it's easier to use and provide guidance on.

I think both are valuable to use but far easier for us to give suggestions with pg_dump.

PR #227, opened 2024-01-19
Add ruby 3.1 to the test matrix

jrafanie:
Load aliases with safe_load if YAML supports it.

Disable YAML.load security cop on versions that don't have safe_load.


jrafanie:
This should be ready to go merge.  ❤️

Issue #229, opened 2024-01-23
Add messaging hostname validation


nasark:
When configuring messaging on an appliance, we should check to see that the hostname is reachable and valid before proceeding with the current configuration. Currently we allow users to enter any hostname and in the event of an incorrect hostname that is entered, the messaging will only fail after the appliance configuration. This can cause unnecessary confusion for users and can be avoided

@miq-bot assign @nasark 
@miq-bot add_label enhancement

nasark:
This is now complete in https://github.com/ManageIQ/manageiq-appliance_console/pull/254

PR #230, opened 2024-01-30
[WIP] Messaging hostname FQDN validation

nasark:
Check if messaging hostname entered matches the FQDN

@miq-bot assign @Fryguy 
@miq-bot add_reviewer @agrare, @jrafanie

Ref: https://github.com/ManageIQ/manageiq-appliance_console/issues/229

### DIFF
--- a/lib/manageiq/appliance_console/message_configuration.rb
+++ b/lib/manageiq/appliance_console/message_configuration.rb
@@ -67,6 +67,8 @@ module ManageIQ
         show_parameters
         return false unless agree("\nProceed? (Y/N): ")
 
+        return false if hostname?(message_server_host) && !hostname_fqdn?(message_server_host)
+
         return false unless host_reachable?(message_server_host, "Message Server Host:")
 
         true
@@ -174,6 +176,21 @@ module ManageIQ
         end
 
         say("Content already exists in #{path}. Taking no action.")
+        true
+      end
+
+      def hostname?(host)
+        host !~ IP_REGEXP
+      end
+
+      def hostname_fqdn?(host)
+        require "socket"
+
+        fqdn = Addrinfo.getaddrinfo(Socket.gethostname, nil).first.getnameinfo.first
+        unless host == fqdn
+          return false
+        end
+
         true
       end
 


lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -177,6 +181,22 @@ def file_contains?(path, content)
...
+      def is_hostname?(host)
+        require "resolv"
+
+        return !(host =~ IP_REGEXP)
+      end
+
+      def hostname_fqdn?(host)
+        require "socket"
+
+        fqdn = Addrinfo.getaddrinfo(Socket.gethostname, nil).first.getnameinfo.first
```
nasark, line 193: Not sure if its better to use `hostname -f` (with back ticks) here?

agrare: What is the ` Addrinfo.getaddrinfo` doing for us?  Can you just use what `Socket.gethostname` returns?

agrare:
@nasark where does the requirement that the messaging server's hostname match what the user enters?
Machines can have multiple hostnames and what a user uses to access a messaging server from other clients might not be what `hostname -f` returns.
And if we _have_ to have them match why have the user enter it at all just to fail if it matches, couldn't we just pre-populate the value?

lib/manageiq/appliance_console/message_configuration.rb
nasark: `Addrinfo.getaddrinfo` provides more info such as the FQDN whereas `Socket.gethostname` will provide the hostname but not necessarily the FQDN. I've tested this on an AWS appliance i.e.
```
irb(main):002:0> Addrinfo.getaddrinfo(Socket.gethostname, nil).first.getnameinfo.first
=> "<host>.us-west-1.compute.internal"
irb(main):003:0> Socket.gethostname
=> "<host>"
```

jrafanie: Are we guaranteed to have the `addrinfo` we want "first" in the array and the `nameinfo` we want to also be first?

Googling around, I found my buddy from the old sql server adapter days, metaskills opened https://github.com/newrelic/newrelic-ruby-agent/issues number 697.  The newrelic ruby agent now defaults to trying `hostname -f` first, then `hostname`, and then heroku dyno name, and finally `gethostname`.  See: https://github.com/newrelic/newrelic-ruby-agent/pull/1015/files

And the current code in main:
https://github.com/newrelic/newrelic-ruby-agent/blob/0824b224b9c1be974d0b47ae9422f2f996046218/lib/new_relic/agent/hostname.rb#L11-L38

nasark: This is a good suggestion. Thoughts on something like this (first time using awesome_spawn)?
```suggestion
        begin
          fqdn = AwesomeSpawn.run!("hostname", :params => ["-f"])
        rescue AwesomeSpawn::CommandResultError
          require "socket"

          fqdn = Addrinfo.getaddrinfo(Socket.gethostname, nil).first.getnameinfo.first
        end

```

jrafanie: I think we'll be able to guarantee if we'll have `hostname` binary installed on appliances.  I guess my question is can we just assume `hostname` or do we need the fallback?  

If we need the fallback, do we need to worry about the order of the arrays (assumption on the first element being the correct one).  I recall some ruby stdlib for ip addresses would sometimes return loopback first instead of your publicly visible ip address so I wasn't sure if we have similar problems if there are multiple results.

agrare: Before we do anything I'd like to understand how we got different values in the messaging yml and the cert.

nasark:
Marking as WIP. After a discussion with @agrare we think its better to solve the problem at a lower level and add validation to prevent scenarios where the host in messaging.yml does not match the host being used in the certs

Issue #232, opened 2024-02-01
Configuring internal database with appliance_console fails


agrare:
```

Activating the configuration using the following settings...
Host:     localhost
Username: root
Database: vmdb_production
Region:   0

Initialize postgresql disk starting
Initialize postgresql disk complete
Initialize postgresql starting
Initialize postgresql complete
Configuration failed: Unknown alias: base

Press any key to continue.
```

I don't see anything interesting in the appliance_console.log
```
# cat log/appliance_console.log 
# Logfile created on 2024-02-01 14:31:10 -0500 by logger.rb/v1.5.0
I, [2024-02-01T14:31:10.224542 #6622]  INFO -- : MIQ(ManageIQ::ApplianceConsole::InternalDatabaseConfiguration#initialize_postgresql_disk) : starting
I, [2024-02-01T14:31:10.846971 #6622]  INFO -- : MIQ(ManageIQ::ApplianceConsole::InternalDatabaseConfiguration#initialize_postgresql_disk) : complete
I, [2024-02-01T14:31:10.855250 #6622]  INFO -- : MIQ(ManageIQ::ApplianceConsole::InternalDatabaseConfiguration#initialize_postgresql) : starting
I, [2024-02-01T14:31:12.887664 #6622]  INFO -- : MIQ(ManageIQ::ApplianceConsole::InternalDatabaseConfiguration#initialize_postgresql) : complete
```

But I'm pretty sure this is YAML safe_load related

agrare:
cc @jrafanie

agrare:
Ah, got the backtrace once I hit the any key
```
/opt/manageiq/manageiq-gemset/gems/psych-4.0.6/lib/psych/visitors/to_ruby.rb:432:in `visit_Psych_Nodes_Alias': Unknown alias: base (Psych::BadAlias)
	from /opt/manageiq/manageiq-gemset/gems/psych-4.0.6/lib/psych/visitors/visitor.rb:30:in `visit'
	from /opt/manageiq/manageiq-gemset/gems/psych-4.0.6/lib/psych/visitors/visitor.rb:6:in `accept'
	from /opt/manageiq/manageiq-gemset/gems/psych-4.0.6/lib/psych/visitors/to_ruby.rb:35:in `accept'
	from /opt/manageiq/manageiq-gemset/gems/psych-4.0.6/lib/psych/visitors/to_ruby.rb:347:in `block in revive_hash'
[...]

PR #234, opened 2024-02-01
Fix YAML.load_file failing on aliases

agrare:
Fix configuring an internal database with ruby 3.1 failing due to aliases in the `config/database.pg.yml` file

```
/opt/manageiq/manageiq-gemset/gems/psych-4.0.6/lib/psych/visitors/to_ruby.rb:432:in 'visit_Psych_Nodes_Alias': Unknown alias: base (Psych::BadAlias)
```

Fixes https://github.com/ManageIQ/manageiq-appliance_console/issues/232

### DIFF
--- a/lib/manageiq/appliance_console/database_configuration.rb
+++ b/lib/manageiq/appliance_console/database_configuration.rb
@@ -257,7 +257,13 @@ FRIENDLY
         require 'fileutils'
         FileUtils.cp(DB_YML_TMPL, DB_YML) if File.exist?(DB_YML_TMPL)
       end
-      YAML.load_file(DB_YML)
+
+      data = File.read(DB_YML)
+      if YAML.respond_to?(:safe_load)
+        YAML.safe_load(data, :aliases => true)
+      else
+        YAML.load(data) # rubocop:disable Security/YAMLLoad
+      end
     end
 
     def validate_encryption_key!


lib/manageiq/appliance_console/database_configuration.rb
```diff
@@ -257,7 +257,12 @@ def self.load_current
         require 'fileutils'
         FileUtils.cp(DB_YML_TMPL, DB_YML) if File.exist?(DB_YML_TMPL)
       end
-      YAML.load_file(DB_YML)
+
+      if YAML.respond_to?(:safe_load_file)
+        YAML.safe_load_file(DB_YML, :aliases => true)
```
agrare, line 262: @jrafanie I noticed in https://github.com/ManageIQ/manageiq-appliance_console/pull/227 you replaced `YAML.load_file` with a `File.read` + `YAML.safe_load`, it looks like `YAML.safe_load_file` accepts `:aliases => true` as well was there a reason you didn't go this route?

agrare: :scream: okay I'll match what you have in messaging_configuration

jrafanie: lol, I kept digging, kwarg for `aliases` was added in 3.1.0: https://github.com/ruby/psych/commit/4d4439d6d0adfcbd211ea295779315f1baa7dadd.  I made the assumption we'd be at 3.1.0 at least so we'll have `safe_load` with kwargs but perhaps not yet have `safe_load_file` or `unsafe_load`/`unsafe_load_file`.

Some gems support super old rubies so it was easier to just follow that pattern.  Technically, it wouldn't work with psych 3 or earlier but 🤷

jrafanie (approved):
LGTM, complexity is failing but this method is not even that bad.  🤣

PR #238, opened 2024-03-04
Remove network configuration from the appliance console

bdunne:
- This is better configured from cockpit, nmcli or nmtui
- The appliance console network configuration is broken on EL9

Related: https://github.com/ManageIQ/manageiq-appliance/pull/381


Fryguy (approved):

Fryguy:
@bdunne After merge how do we want to release this?  What are the next steps for documentation? (I presume there's some deletes and then some new pages we need to create?)

bdunne:
>After merge how do we want to release this? 

I would assume this is a major version since it's a breaking change.

>  What are the next steps for documentation? (I presume there's some deletes and then some new pages we need to create?)

I spent some time looking through the documentation, but don't see anything related to this to change there.
Maybe it should be part of our release announcement?

PR #244, opened 2024-04-12
Fix post_activation exception when setting up internal database

agrare:
Once both database and messaging have been configured by the CLI, start evmserverd.service.  This will work if you pass both database and messaging config in the same invocation or if you configure them separately.

Example command: `appliance_console_cli --internal --username='root' --password='smartvm' --region=1 --dbdisk=/dev/vdb --message-server-config --message-keystore-password="smartvm" --server=start`

Fixes:
```
  config.post_activation
        ^^^^^^^^^^^^^^^^
from /opt/manageiq/manageiq-gemset/gems/manageiq-appliance_console-9.0.1/lib/manageiq/appliance_console/cli.rb:297:in `set_db'
from /opt/manageiq/manageiq-gemset/gems/manageiq-appliance_console-9.0.1/lib/manageiq/appliance_console/cli.rb:261:in `run'
from /opt/manageiq/manageiq-gemset/gems/manageiq-appliance_console-9.0.1/lib/manageiq/appliance_console/cli.rb:544:in `parse'
from /opt/manageiq/manageiq-gemset/gems/manageiq-appliance_console-9.0.1/bin/appliance_console_cli:7:in `<top (required)>'
from /opt/manageiq/manageiq-gemset/bin/appliance_console_cli:25:in `load'
from /opt/manageiq/manageiq-gemset/bin/appliance_console_cli:25:in `<main>'
```

Introduced by https://github.com/ManageIQ/manageiq-appliance_console/pull/195

Follow-up:
* https://github.com/ManageIQ/manageiq-documentation/pull/1789

### DIFF
--- a/lib/manageiq/appliance_console/cli.rb
+++ b/lib/manageiq/appliance_console/cli.rb
@@ -273,12 +273,14 @@ module ApplianceConsole
       saml_unconfig if saml_unconfig?
       oidc_config if oidc_config?
       oidc_unconfig if oidc_unconfig?
-      set_server_state if set_server_state?
       openscap if openscap?
       message_server_config if message_server_config?
       message_server_unconfig if message_server_unconfig?
       message_client_config if message_client_config?
       message_client_unconfig if message_client_unconfig?
+      # set_server_state must be after set_db and message_*_config so that a user
+      # can configure database, messaging, and start the server in one command
+      set_server_state if set_server_state?
     rescue CliError => e
       say(e.message)
       say("")
@@ -322,9 +324,6 @@ module ApplianceConsole
       # start pg, create user, create db update the rails configuration,
       # verify, set up the database with region. activate does it all!
       raise CliError, "Failed to configure internal database" unless config.activate
-
-      # enable/start related services
-      config.post_activation
     rescue RuntimeError => e
       raise CliError, "Failed to configure internal database #{e.message}"
     end
@@ -343,9 +342,6 @@ module ApplianceConsole
 
       # call create_or_join_region (depends on region value)
       raise CliError, "Failed to configure external database" unless config.activate
-
-      # enable/start related services
-      config.post_activation
     end
 
     def set_replication


agrare:
Unsure if it is related or not but I have to manually run `db:migrate db:seed` in order for 
```
[root@manageiq-devel ~]# appliance_console_cli --internal --username='root' --password='smartvm' --region=1 --dbdisk=/dev/vdb --standalone --message-server-config --message-keystore-password="smartvm"
create encryption key
configuring internal database
Initialize postgresql disk starting
Initialize postgresql disk complete
Initialize postgresql starting
Initialize postgresql complete
Create Jaas Config
Create Client Properties
Create Logs Directory
Configure Firewall
Configure Keystore
Create Server Properties
Configure Messaging Yaml
Starting zookeeper and configure it to start on reboots ...
Starting kafka and configure it to start on reboots ...
```

to complete.  Looking into this.

lib/manageiq/appliance_console/cli.rb
```diff
@@ -273,12 +273,12 @@ def run
       saml_unconfig if saml_unconfig?
       oidc_config if oidc_config?
       oidc_unconfig if oidc_unconfig?
-      set_server_state if set_server_state?
       openscap if openscap?
       message_server_config if message_server_config?
       message_server_unconfig if message_server_unconfig?
       message_client_config if message_client_config?
       message_client_unconfig if message_client_unconfig?
+      set_server_state if set_server_state?
```
agrare, line 281: NOTE moving this after the `message_*_config` so that you can configure db and/or messaging and also start the server

Fryguy: Can you add that as a literal comment so someone doesn't come along in the future and just reorder them (say, for like alphabetical reasons)?  Asking because I don't think we have specs here.

Fryguy:
I like the single step version and we can take care of the dependencies internally.  So, yeah, seems we can just document that.

agrare:
Working on adding documentation for the messaging config since it seems that wasn't added, I'll add the `--server=start` to the examples so people should copy&paste the right thing

lib/manageiq/appliance_console/cli.rb
agrare: :+1: done

agrare:
Okay I'm going to take this out of WIP since this is functional and the only issues left aren't related to this fix (specifically https://github.com/ManageIQ/manageiq-appliance_console/issues/245)

Fryguy (approved):

Issue #245, opened 2024-04-15
Unable to fetch key from remote system, fails with `dh#set_pqg= is incompatible with OpenSSL 3.0`


agrare:
```
[root@manageiq-1 ~]# appliance_console_cli --fetch-key manageiq-0.localdomain --sshlogin root --sshpassword smartvm
fetch encryption key
Failed to fetch key: dh#set_pqg= is incompatible with OpenSSL 3.0
Could not create encryption key (v2_key)
```

```
Configure Application

No encryption key found.
For migrations, copy encryption key from a hardened appliance.
For worker and multi-region setups, copy key from another appliance.
If this is your first appliance, just generate one now.

Encryption Key

1) Create key
2) Fetch key from remote machine
(1) 
Choose the encryption key: |1| 2

Enter the hostname for appliance with encryption key: manageiq-0.localdomain
Enter the appliance SSH login: |root|
Enter the appliance SSH password: *******
Enter the path of remote encryption key: |/var/www/miq/vmdb/certs/v2_key|
Failed to fetch key: dh#set_pqg= is incompatible with OpenSSL 3.0/v2_key
Try again? (Y/N)
```

Backtrace:
```
"/home/grare/adam/.gem/ruby/3.1.0/gems/net-ssh-4.2.0/lib/net/ssh/buffer.rb:276:in `set_key'",
 "/home/grare/adam/.gem/ruby/3.1.0/gems/net-ssh-4.2.0/lib/net/ssh/buffer.rb:276:in `read_keyblob'",
 "/home/grare/adam/.gem/ruby/3.1.0/gems/net-ssh-4.2.0/lib/net/ssh/buffer.rb:248:in `read_key'",
 "/home/grare/adam/.gem/ruby/3.1.0/gems/net-ssh-4.2.0/lib/net/ssh/known_hosts.rb:154:in `block (2 levels) in keys_for'",
 "/home/grare/adam/.gem/ruby/3.1.0/gems/net-ssh-4.2.0/lib/net/ssh/known_hosts.rb:136:in `each_line'",
 "/home/grare/adam/.gem/ruby/3.1.0/gems/net-ssh-4.2.0/lib/net/ssh/known_hosts.rb:136:in `block in keys_for'",
 "/home/grare/adam/.gem/ruby/3.1.0/gems/net-ssh-4.2.0/lib/net/ssh/known_hosts.rb:134:in `open'",
[...]

PR #248, opened 2024-04-16
Bump net-ssh/scp for OpenSSL 3.0 support

agrare:
Fix `dh#set_pqg= is incompatible with OpenSSL 3.0`

https://github.com/net-ssh/net-ssh/pull/864

Depends on:
- [x] https://github.com/ManageIQ/linux_admin/pull/244
- [x] Release of linux_admin including :point_up: 

Fixes https://github.com/ManageIQ/manageiq-appliance_console/issues/245


manageiq-appliance_console.gemspec
```diff
@@ -25,11 +25,13 @@ Gem::Specification.new do |spec|
...
   spec.add_runtime_dependency "bcrypt",                  "~> 3.1.10"
   spec.add_runtime_dependency "bcrypt_pbkdf",            ">= 1.0", "< 2.0"
+  spec.add_runtime_dependency "ed25519",                 ">= 1.2", "< 1.3"
   spec.add_runtime_dependency "highline",                "~> 2.1"
   spec.add_runtime_dependency "i18n",                    ">= 0.8"
   spec.add_runtime_dependency "linux_admin",             "~> 2.0"
   spec.add_runtime_dependency "manageiq-password",       "< 2"
-  spec.add_runtime_dependency "net-scp",                 "~> 1.2.1"
+  spec.add_runtime_dependency "net-scp",                 "~> 4.0"
+  spec.add_runtime_dependency "net-ssh",                 "~>7.2"
```
Fryguy, line 34:
```suggestion
  spec.add_runtime_dependency "net-ssh",                 "~> 7.2"
```

manageiq-appliance_console.gemspec
```diff
@@ -25,11 +25,13 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "awesome_spawn",           "~> 1.6"
   spec.add_runtime_dependency "bcrypt",                  "~> 3.1.10"
   spec.add_runtime_dependency "bcrypt_pbkdf",            ">= 1.0", "< 2.0"
+  spec.add_runtime_dependency "ed25519",                 ">= 1.2", "< 1.3"
```
Fryguy, line 28: Out of curiosity why is this explicitly needed?

manageiq-appliance_console.gemspec
Fryguy, line 34: @bdunne Please review - I'm curious if there are follow on effect of this large of a bump in both of these.

manageiq-appliance_console.gemspec
agrare:
```
E, [2024-04-16T13:41:31.316966 #670326] ERROR -- net.ssh.authentication.agent[b18]: ignoring unimplemented key:unsupported key type `ssh-ed25519'
net-ssh requires the following gems for ed25519 support:
 * ed25519 (>= 1.2, < 2.0)
 * bcrypt_pbkdf (>= 1.0, < 2.0)
See https://github.com/net-ssh/net-ssh/issues/565 for more information
Gem::LoadError : "ed25519 is not part of the bundle. Add it to your Gemfile."
```

manageiq-appliance_console.gemspec
agrare: Done

manageiq-appliance_console.gemspec
Fryguy: ohhh interesting - wonder if we can drop the rbnacl gem then?

Fryguy: See https://github.com/ManageIQ/manageiq-appliance_console/commit/67653544afebe7bdbfedeed9c9161cac38592aeb - I think they were only added for this specific cipher.

Getting rid of those would be huge cause they require a C lib.

agrare: Added in 67653544afebe7bdbfedeed9c9161cac38592aeb

Looks like yes, that was the only reason rbnacl was added

We might be able to drop rbnacl-devel rpm also: 350c879fe1f080bf154bea3e0616d51b0c0d9ee6

agrare: I think we're okay actually, `jwt` just needs libsodium not `rbnacl`: https://github.com/ManageIQ/manageiq-rpm_build/pull/117#issuecomment-724800717

Fryguy: yeah it probably should have been there all along in both places.

agrare: Agreed, https://github.com/ManageIQ/manageiq-providers-google/pull/264

agrare: Removed here https://github.com/ManageIQ/manageiq-appliance_console/pull/248/files#diff-5cfba543196ff7509ccbd7b987182595a8baa8cc3e4a4e1040e20ccfd75d885eL36

Fryguy (approved):
LGTM but would like a second reviewer... @bdunne ?

manageiq-appliance_console.gemspec
```diff
@@ -25,15 +25,16 @@ Gem::Specification.new do |spec|
...
```
bdunne, line 28: Is there a problem with 1.3?

agrare: Ah 1.3 was what they had in the description of the linked issue https://github.com/net-ssh/net-ssh/issues/565 but the actual documentation does have `< 2.0` I'll update to use that instead.

agrare: Done

bdunne:
LinuxAdmin v3.0.0 is available https://rubygems.org/gems/linux_admin/versions/3.0.0

PR #251, opened 2024-05-10
Add a common method for asking for a new password

agrare:
When entering an existing password it is sufficient to prompt once since if the user mistyped the password it will fail later on, but when creating a new password it is common to re-prompt them to confirm that they didn't typo the new password.

We had this "prompt twice, check non-empty" loop in two places and were going to need it in a third so it seemed better to move this to a common method.

TODO
- [x] Live test on an appliance

Database configuration testing:
```
Each database region number must be unique.
Enter the database region number: 0
Enter the database password on localhost: 

Password can not be empty, please try again
Enter the database password on localhost: 

Password can not be empty, please try again
Enter the database password on localhost: *******
Enter the Re-enter database password on localhost: *****

The passwords did not match, please try again
Enter the database password on localhost: *******
Enter the Re-enter database password on localhost: *******
```

Message Server Configuration testing
```
Installed file /opt/kafka/config/server.properties found.

Already configured on this Appliance, Un-Configure first? (Y/N): Y
Remove Installed Files
Unconfigure Firewall
Deactivate Services

Proceed with Configuration? (Y/N): Y

Message Server Parameters:

Enter the Message Keystore Password: |admin| ss: |manageiq-devel.localdomain| 

Password can not be empty, please try again
Enter the Message Keystore Password: *******
Enter the Re-enter Message Keystore Password: *****

The passwords did not match, please try again
Enter the Message Keystore Password: *******
Enter the Re-enter Message Keystore Password: *******
```

Fixes https://github.com/ManageIQ/manageiq/issues/23031

### DIFF
--- a/lib/manageiq/appliance_console/database_replication.rb
+++ b/lib/manageiq/appliance_console/database_replication.rb
@@ -6,6 +6,7 @@ module ManageIQ
 module ApplianceConsole
   class DatabaseReplication
     include ManageIQ::ApplianceConsole::Logging
+    include ManageIQ::ApplianceConsole::Prompts
 
     PGPASS_FILE       = '/var/lib/pgsql/.pgpass'.freeze
     NETWORK_INTERFACE = 'eth0'.freeze
@@ -115,25 +116,9 @@ Replication Server Configuration
     private
 
     def ask_for_cluster_database_credentials
-      self.database_name = just_ask("cluster database name", database_name)
-      self.database_user = just_ask("cluster database username", database_user)
-
-      count = 0
-      loop do
-        count += 1
-        password1 = ask_for_password("cluster database password", database_password)
-        # if they took the default, just bail
-        break if password1 == database_password
-        password2 = ask_for_password("cluster database password")
-        if password1 == password2
-          self.database_password = password1
-          break
-        elsif count > 1 # only reprompt password once
-          raise RuntimeError, "passwords did not match"
-        else
-          say("\nThe passwords did not match, please try again")
-        end
-      end
+      self.database_name     = just_ask("cluster database name", database_name)
+      self.database_user     = just_ask("cluster database username", database_user)
+      self.database_password = ask_for_new_password("cluster database password", :default => database_password)
     end
 
     def run_repmgr_command(cmd, params = {})
--- a/lib/manageiq/appliance_console/prompts.rb
+++ b/lib/manageiq/appliance_console/prompts.rb
@@ -113,6 +113,27 @@ module ApplianceConsole
       pass == "********" ? (default || "") : pass
     end
 
+    def ask_for_new_password(prompt, default: nil, allow_empty: false, retry_limit: 1, confirm_password: true)
+      count = 0
+      loop do
+        password1 = ask_for_password(prompt, default)
+        if password1.strip.empty? && !allow_empty
+          say("\nPassword can not be empty, please try again")
+          next
+        end
+
+        return password1 if password1 == default || !confirm_password
+
+        password2 = ask_for_password(prompt)
+        return password1 if password1 == password2
+
+        raise "passwords did not match" if count >= retry_limit
+
+        count += 1
+        say("\nThe passwords did not match, please try again")
+      end
+    end
+
     def ask_for_string(prompt, default = nil)
       just_ask(prompt, default)
     end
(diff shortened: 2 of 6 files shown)


lib/manageiq/appliance_console/database_replication.rb
```diff
@@ -115,25 +116,9 @@ def write_pgpass_file
...
-          break
-        elsif count > 1 # only reprompt password once
-          raise RuntimeError, "passwords did not match"
-        else
-          say("\nThe passwords did not match, please try again")
-        end
-      end
+      self.database_name     = just_ask("cluster database name", database_name)
+      self.database_user     = just_ask("cluster database username", database_user)
+      self.database_password = ask_for_new_password("cluster database password", :default => database_password, :allow_empty => true)
```
agrare, line 121: NOTE I'm not sure if this was intentional but the old method did allow for an empty password

Fryguy: That's weird - @jrafanie or @bdunne do you know?

Fryguy: Or maybe because it has a default, allow empty allows it to use the default?

lib/manageiq/appliance_console/prompts.rb
```diff
@@ -113,6 +113,27 @@ def ask_for_password(prompt, default = nil)
...
+      loop do
+        password1 = ask_for_password(prompt, default)
+        if password1.strip.empty? && !allow_empty
+          say("\nPassword can not be empty, please try again")
+          next
+        end
+
+        return password1 if password1 == default || !confirm_password
+
+        password2 = ask_for_password("#{prompt} again")
```
Fryguy, line 127: I know this "again" came from one of the examples, but I think this would be a nicer presentation:

```suggestion
        password2 = ask_for_password("Re-enter #{prompt}")
```

agrare: :+1: changed

lib/manageiq/appliance_console/database_replication.rb
agrare: https://github.com/ManageIQ/manageiq-appliance_console/pull/251/files#diff-58f7d485041385eb768bd8d25d7e83a11ccf1a1bbf2bc4fb136fcdcf9cd6b590L126 had a default as well but that checks for a 0 length password.  I think the default is only present if you've already configured a password but if this is the first time through there would be no default.

jrafanie: Looks like it was like that back to the initial open source commit.

It had a concept of no password, which I assume means the default, likely already configured with a password.  I'm not sure though.


```diff
+    def ask_for_database_credentials
+      self.host     = ask_for_ip_or_hostname("database hostname or IP address", host) if host.blank? || !local?
+      self.database = just_ask("name of the database on #{host}", database) unless local?
+      self.username = just_ask("username", username) unless local?
+      count = 0
+      loop do
+        count += 1
+        password1   = ask_for_password_or_none("database password on #{host}", password)
+        # if they took the default, just bail
+        break if (password1 == password)
+        password2   = ask_for_password("database password again")
+        if password1 == password2
+          self.password = password1
+          break
+        elsif count > 1 # only reprompt password once
+          raise ArgumentError, "passwords did not match"
+        else
+          say("\nThe passwords did not match, please try again")
+        end
+      end
+    end
```

af467356d71d5c81990590933254cd54cde2c6b7

agrare: Okay I think I'm going to keep the behavior the same as the internal database configuration and not allow a non-default empty password.

Fryguy: @agrare I agree :+1:

lib/manageiq/appliance_console/prompts.rb
agrare: @Fryguy `just_ask` prepends `"Enter the "` to the prompt, so when I have it this way we end up with
```
Enter the database password on localhost: *******
Enter the Re-enter database password on localhost: *****
```

Fryguy: Ugh - that's annoying :(

Would prefer if it was 

```
Enter the database password on localhost: *******
Re-enter the database password on localhost: *****
```

If we can't do that it's fine - we can go back to the "again" text - sorry for the runaround.

agrare: Yeah so the cluster database just printed the same string twice, prefer that or the "again" suffix?
Longer term I started working on just_ask taking a prefix argument which defaults to "Enter the " but we could replace that with Re-enter.  It is all positional arguments right now though so need to check all of the callers and make sure I don't break something

Fryguy: Yeah I was thinking of a parameter to just_ask also...can definitely be a followup.  I think putting the "#{original_text} again" text is fine for this PR.

PR #254, opened 2024-07-17
Messaging Hostname Validation

nasark:
- reject IPs and localhost entries when prompted for messaging hostname
- check if hostname is resolvable alongside reachability after prompts

@miq-bot assign @agrare 
@miq-bot add_reviewer @agrare 

### DIFF
--- a/lib/manageiq/appliance_console/message_configuration.rb
+++ b/lib/manageiq/appliance_console/message_configuration.rb
@@ -68,7 +68,7 @@ module ManageIQ
         show_parameters
         return false unless agree("\nProceed? (Y/N): ")
 
-        return false unless host_reachable?(message_server_host, "Message Server Host:")
+        return false unless host_resolvable?(message_server_host) && host_reachable?(message_server_host, "Message Server Host:")
 
         true
       end
@@ -186,6 +186,29 @@ module ManageIQ
           say("the #{what} must be reachable by name.")
           return false
         end
+        say("Succeeded.")
+        true
+      end
+
+      def host_resolvable?(host)
+        require 'ipaddr'
+        require 'resolv'
+
+        say("Checking if #{host} is resolvable ... ")
+        begin
+          ip_address = Resolv.getaddress(host)
+
+          if IPAddr.new("127.0.0.1/8").include?(ip_address) || IPAddr.new("::1/128").include?(ip_address)
+            say("Failed.\nThe hostname must not resolve to a link-local address")
+
+            return false
+          end
+        rescue Resolv::ResolvError => e
+          say("Failed.\nHostname #{host} is not resolvable: #{e.message}")
+
+          return false
+        end
+
         say("Succeeded.")
         true
       end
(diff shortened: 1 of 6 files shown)


lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -190,6 +190,27 @@ def host_reachable?(host, what)
         true
       end
 
+      def host_resolvable?(host)
+        require 'resolv'
+
+        say("Checking if #{host} is resolvable ... ")
+        begin
+          ip_address = Resolv.getaddress(host)
+          if ip_address == "127.0.0.1" || ip_address == "::1"
```
agrare, line 199: This is probably good enough, technically a link local address is a whole range of addrs `127.0.0.1/8` / `::1/128`

We could use IPAddr to do the subnet check, e.g.`IPAddr.new("127.0.0.1/8") === IPAddr.new("127.0.1.2")`

agrare:
```suggestion
          if IPAddr.new("127.0.0.1/8").include?(ip_address) || IPAddr.new("::1/128").include?(ip_address)
```

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -190,6 +190,27 @@ def host_reachable?(host, what)
...
+            say("Failed.\nThe hostname must not resolve to 127.0.0.1")
```
agrare, line 200:
```suggestion
            say("Failed.\nThe hostname must not resolve to a link-local address")
```

nasark:
updated specs, removing WIP

lib/manageiq/appliance_console/message_configuration.rb
```diff
@@ -190,6 +190,28 @@ def host_reachable?(host, what)
         true
       end
 
+      def host_resolvable?(host)
+        require 'resolv'
+
```
agrare, line 195: @nasark can you `require 'ipaddr'` here as well just to be safe?

lib/manageiq/appliance_console/prompts.rb
```diff
@@ -71,6 +72,11 @@ def ask_for_hostname(prompt, default = nil, validate = HOSTNAME_REGEXP, error_te
       just_ask(prompt, default, validate, error_text, &block)
     end
 
+    def ask_for_messaging_hostname(prompt, default = nil, error_text = "a valid Messaging Hostname (not an IP or localhost)", &block)
+      validation = ->(h) { h =~ MESSAGING_HOSTNAME_REGEXP && h !~ IP_REGEXP }
```
agrare, line 76: Hm I wonder if instead of mostly duplicating the hostname regex we just say `h =~ HOSTNAME_REGEXP && h !~ IP_REGEXP && h != "localhost"` here?

nasark: not a bad idea, the only difference I see is that the regex detects instances of `localhost` anywhere in the string. i.e. a hostname that has `.localhost` at the end (or anywhere), another example is the infamous `localhost.localadmin`

agrare (approved):

PR #255, opened 2024-07-22
Messaging keystore password validation

nasark:
During appliance install, when using a messaging keystore password with special characters the appliance setup will fail. This appears to be a JKS (Java Keystore) limitation and should be addressed by preventing entry of passwords with special characters

@miq-bot assign @agrare 
@miq-bot add_label bug
@miq-bot add_reviewer @agrare


agrare:
@nasark does JKS not support _any_ special characters or only some?  I've used passwords before with some special characters and they appear to have worked, I know in this specific case it was `#` that failed

nasark:
> @nasark does JKS not support any special characters or only some? I've used passwords before with some special characters and they appear to have worked, I know in this specific case it was # that failed

@agrare The only special characters that worked in my testing are `$`, `*`, which I can add to the regex to allow them... it's possible that the allowed special characters differ between Java versions

agrare:
Nope those just happened to be the two special chars that I used so I thought maybe it was only `#` that failed :)

Yeah could you add those to the regex and update the error message to indicate what is allowed

lib/manageiq/appliance_console/prompts.rb
```diff
@@ -140,6 +146,10 @@ def ask_for_new_password(prompt, default: nil, allow_empty: false, retry_limit:
       end
     end
 
+    def ask_for_messaging_password(prompt)
+      ask_for_new_password(prompt, :validation => MESSAGING_PASSWORD_REGEXP, :validation_err => "special characters are not allowed, except for \"$\" and \"*\"")
```
agrare, line 150: Hm maybe `Allowed characters are a-z, A-Z, 0-9, $, and *`

nasark: thanks, also included `-, _, .,` in the message. Apparently they are not considered to be special characters 😛

agrare: They're all special in their own way 😀

agrare (approved):

PR #257, opened 2024-07-23
Deprecate message-server-use-ipaddr option from CLI

nasark:
We no longer allow IPs to be used for messaging hosts and so the `message-server-use-ipaddr` option in the CLI should be deprecated

@miq-bot assign @agrare 
@miq-bot add_label bug
@miq-bot add_reviewer @agrare

### DIFF
--- a/lib/manageiq/appliance_console/cli.rb
+++ b/lib/manageiq/appliance_console/cli.rb
@@ -216,20 +216,16 @@ module ApplianceConsole
         opt :message_server_username,     "Message Server Username",                                        :type => :string
         opt :message_server_password,     "Message Server password",                                        :type => :string
         opt :message_server_port,         "Message Server Port",                                            :type => :integer
-        opt :message_server_use_ipaddr,   "Message Server Use Address",                                     :type => :boolean, :default => false
+        opt :message_server_use_ipaddr,   "Deprecated: Message Server Use Address",                         :type => :boolean, :default => false
         opt :message_server_host,         "Message Server Hostname or IP Address",                          :type => :string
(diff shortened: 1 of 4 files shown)


agrare:
@nasark could you add a warning if someone passes the ip address flag?  Something like `$stderr.puts "--message_server_use_ipaddr is deprecated and will be ignored" if options[:message_server_use_ipaddr]`

lib/manageiq/appliance_console/cli.rb
```diff
@@ -216,22 +216,18 @@ def parse(args)
...
         opt :message_server_host,         "Message Server Hostname or IP Address",                          :type => :string
         opt :message_truststore_path_src, "Message Server Truststore Path",                                 :type => :string
         opt :message_ca_cert_path_src,    "Message Server CA Cert Path",                                    :type => :string
         opt :message_persistent_disk,     "Message Persistent Disk Path",                                   :type => :string
       end
       Optimist.die :region, "needed when setting up a local database" if region_number_required? && options[:region].nil?
-      Optimist.die "Supply only one of --message-server-host or --message-server-use-ipaddr=true" if both_host_and_use_ip_addr_specified?
       Optimist.die "Supply only one of --message-server-config, --message-server-unconfig, --message-client-config or --message-client-unconfig" if multiple_message_subcommands?
+      $stderr.puts "Warning: --message_server_use_ipaddr is deprecated and will be ignored" if options[:message_server_use_ipaddr]
```
agrare, line 227: Rubocop has a good point, warn would be better my bad
```suggestion
      warn("--message_server_use_ipaddr is deprecated and will be ignored") if options[:message_server_use_ipaddr]
```

agrare (approved):

PR #263, opened 2024-08-06
Don't depend on eth0 for displaying network information

agrare:
When displaying the current network information iterate over all network interfaces not just eth0

Example output:
```
Welcome to the ManageIQ Virtual Appliance.

To modify the configuration, use a web browser to access the management page.

Hostname:                workstation
enp193s0f0np0 IPv4 Address: 10.2.4.10/24
enp193s0f0np0 IPv4 Gateway: 10.2.4.1
enp193s0f0np0 IPv6 Address: 
enp193s0f0np0 IPv6 Gateway: 
enp193s0f0np0 MAC Address: b4:96:91:b8:a2:96
lxcbr0 IPv4 Address:     10.0.3.1/24
lxcbr0 IPv4 Gateway:     10.2.4.1
lxcbr0 IPv6 Address:     fc11:4514:1919:810::1/64
lxcbr0 IPv6 Gateway:     
lxcbr0 MAC Address:      00:16:3e:00:00:00
crc IPv4 Address:        192.168.130.1/24
crc IPv4 Gateway:        10.2.4.1
crc IPv6 Address:        
crc IPv6 Gateway:        
crc MAC Address:         52:54:00:fd:be:d0
Primary DNS:             127.0.0.1
Secondary DNS:           
Search Order:            rb.nj.grare.com
Timezone:                America/New_York
ManageIQ Server:         not configured
ManageIQ Database:       not configured
Database/Region:         not configured
Messaging:               configured
Local Messaging Broker:  configured
ManageIQ Version: 
```

NOTE with longer interface names the spacing is broken

Related: https://github.com/ManageIQ/manageiq-appliance-build/pull/573

### DIFF
--- a/lib/manageiq/appliance_console/database_replication_primary.rb
+++ b/lib/manageiq/appliance_console/database_replication_primary.rb
@@ -10,7 +10,7 @@ module ApplianceConsole
       self.database_name     = "vmdb_production"
       self.database_user     = "root"
       self.database_password = nil
-      self.primary_host      = LinuxAdmin::NetworkInterface.new(NETWORK_INTERFACE).address
+      self.primary_host      = network_interfaces.first&.address
     end
 
     def ask_questions
--- a/bin/appliance_console
+++ b/bin/appliance_console
@@ -72,13 +72,13 @@ RE_RESTART  = "Restart".freeze
 RE_DELLOGS  = "Restart and Clean Logs".freeze
 RE_OPTIONS  = [RE_RESTART, RE_DELLOGS, ManageIQ::ApplianceConsole::CANCEL].freeze
 
-NETWORK_INTERFACE = "eth0".freeze
 CLOUD_INIT_NETWORK_CONFIG_FILE = "/etc/cloud/cloud.cfg.d/99_miq_disable_network_config.cfg".freeze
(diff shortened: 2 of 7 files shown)


agrare:
I thought this would help put some of the changes to LinuxAdmin in context of how they might be used

lib/manageiq/appliance_console/database_replication_primary.rb
```diff
@@ -10,7 +10,7 @@ def initialize
       self.database_name     = "vmdb_production"
       self.database_user     = "root"
       self.database_password = nil
-      self.primary_host      = LinuxAdmin::NetworkInterface.new(NETWORK_INTERFACE).address
+      self.primary_host      = network_interfaces.first&.address
```
agrare, line 13: NOTE this will only work reliably with a single non-loopback interface, but this is just the default presented to the user and they can enter any address that they want plus our appliances only ship by default with a single NIC.

In the future I think a `#default?` attribute or something like that which checks which dev is used for the default route would let us pick the interface most likely to be the "right one" :tm:

bdunne: I think that's fine.  There was no guarantee before that eth0 was the default route.

PR #265, opened 2024-08-23
CP4AIOPS-3113 Store saml remote user configuration separately from sssd lookup

kbrock:
Part of https://github.com/ManageIQ/manageiq/pull/23139

saml configuration (mod_auth_mellon) uses a different delimiter from sssd type configurations.

The templates have been changed to reflect this change. The appliance console is now respecting this change

### DIFF
--- a/lib/manageiq/appliance_console/saml_authentication.rb
+++ b/lib/manageiq/appliance_console/saml_authentication.rb
@@ -90,13 +90,15 @@ module ManageIQ
 
       def copy_apache_saml_configfiles
         debug_msg("Copying Apache SAML Config files ...")
-        copy_template(HTTPD_CONFIG_DIRECTORY, "manageiq-remote-user.conf")
+        copy_template(HTTPD_CONFIG_DIRECTORY, "manageiq-remote-user-saml.conf")
         copy_template(HTTPD_CONFIG_DIRECTORY, "manageiq-external-auth-saml.conf")
       end
 
       def remove_apache_saml_configfiles
         debug_msg("Removing Apache SAML Config files ...")
+        # legacy systems may have manageiq-remote-user.conf instead of manageiq-remote-user-saml.conf
         remove_file(HTTPD_CONFIG_DIRECTORY.join("manageiq-remote-user.conf"))
+        remove_file(HTTPD_CONFIG_DIRECTORY.join("manageiq-remote-user-saml.conf"))
         remove_file(HTTPD_CONFIG_DIRECTORY.join("manageiq-external-auth-saml.conf"))
       end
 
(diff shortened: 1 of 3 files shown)


lib/manageiq/appliance_console/saml_authentication.rb
```diff
@@ -90,13 +90,21 @@ def fetch_idp_metadata
...
+        else
+          copy_template(HTTPD_CONFIG_DIRECTORY, "manageiq-remote-user.conf")
+        end
         copy_template(HTTPD_CONFIG_DIRECTORY, "manageiq-external-auth-saml.conf")
       end
 
       def remove_apache_saml_configfiles
         debug_msg("Removing Apache SAML Config files ...")
+        # legacy systems may still have it stored as the old name
         remove_file(HTTPD_CONFIG_DIRECTORY.join("manageiq-remote-user.conf"))
```
Fryguy, line 106: Let's change this comment to be only on the line it comments on.  Otherwise it appears like it applies to the whole "paragraph" of code.

```suggestion
        remove_file(HTTPD_CONFIG_DIRECTORY.join("manageiq-remote-user.conf")) # legacy systems may still have it stored as the old name
```

lib/manageiq/appliance_console/version.rb
```diff
@@ -1,5 +1,5 @@
 module ManageIQ
   module ApplianceConsole
-    VERSION = '9.1.1'.freeze
+    VERSION = '9.2'.freeze
```
Fryguy, line 3:
```suggestion
    VERSION = '9.2.0'.freeze
```

lib/manageiq/appliance_console/version.rb
```diff
@@ -1,5 +1,5 @@
 module ManageIQ
   module ApplianceConsole
-    VERSION = '9.1.1'.freeze
+    VERSION = '10.0'.freeze
```
Fryguy, line 3:
```suggestion
    VERSION = '10.0.0'.freeze
```

lib/manageiq/appliance_console/saml_authentication.rb
kbrock: thnx.
think we're getting away from same line comments (rubocop)
Will change wording to make more clear

kbrock:
update:
- added support for file not present on appliance

update:
- changed version to 10.0
- removed comment about version numbers

update:
- changed version to 10.0.0
- changed comment to better explain why the 2 deletes are present

lib/manageiq/appliance_console/saml_authentication.rb
Fryguy: I disagree with rubocop on that :)

Fryguy (approved):

PR #267, opened 2024-10-23
Fix warnings on newer ERB versions

Fryguy:
Fixes the follow warnings seen in the logs:

```
warning: Passing safe_level with the 2nd argument of ERB.new is deprecated. Do not use it, and specify other arguments as keyword arguments.
warning: Passing trim_mode with the 3rd argument of ERB.new is deprecated. Use keyword argument like ERB.new(str, trim_mode: ...) instead.
```

@jrafanie Please review.

### DIFF
--- a/lib/manageiq/appliance_console/external_httpd_authentication/external_httpd_configuration.rb
+++ b/lib/manageiq/appliance_console/external_httpd_authentication/external_httpd_configuration.rb
@@ -201,7 +201,7 @@ module ApplianceConsole
         src_path  = path_join(src_dir, file)
         dest_path = path_join(dest_dir, file.gsub(".erb", ""))
         if src_path.to_s.include?(".erb")
-          File.write(dest_path, ERB.new(File.read(src_path), nil, '-').result(binding))
+          File.write(dest_path, ERB.new(File.read(src_path), trim_mode: '-').result(binding))
         else
           FileUtils.cp src_path, dest_path
         end


lib/manageiq/appliance_console/external_httpd_authentication/external_httpd_configuration.rb
```diff
@@ -201,7 +201,7 @@ def cp_template(file, src_dir, dest_dir = "/")
         src_path  = path_join(src_dir, file)
         dest_path = path_join(dest_dir, file.gsub(".erb", ""))
         if src_path.to_s.include?(".erb")
-          File.write(dest_path, ERB.new(File.read(src_path), nil, '-').result(binding))
```
jrafanie, line 204: Silly question:  do we use erb?  I'm not sure which path is the common path.

If we need to keep it, I'm fine with either keeping this WIP until we drop ruby 3.0 or add a conditional and remove it after we drop 3.0.  I'm leaning towards just WIP until 3.0 is dropped just because it's unclear how often we'll go through this code path to clean it up afterwards.

Fryguy: Yeah I'm leaning towards waiting on the WIP too.

> Silly question: do we use erb? I'm not sure which path is the common path.

Yes, for example, this is the template for the sssd.conf: https://github.com/ManageIQ/manageiq-appliance/blob/master/TEMPLATE/etc/sssd/sssd.conf.erb

Fryguy:
I'm not sure how, but I was totally wrong - erb 2.2.0 has the keyword args interface, so we should be fine.

jrafanie:
> I'm not sure how, but I was totally wrong - erb 2.2.0 has the keyword args interface, so we should be fine.

Agreed.  As we found in https://github.com/ManageIQ/manageiq/pull/23250, it looks like this is good to go.

lib/manageiq/appliance_console/external_httpd_authentication/external_httpd_configuration.rb
jrafanie: Just, FYI, as found in https://github.com/ManageIQ/manageiq/pull/23250, this is fine... ruby 2.6 was the first one that allowed either trim_mode positional or kwarg.  We can safely remove the safe_level too so it's good to go

jrafanie (approved):

PR #270, opened 2024-11-26
readline does not work with ruby 3.3

bdunne:
readline is making the prompt disappear when enabled, disable for now until we can get to highline 3.0+ where they switch to reline.

Unfortunately this breaks the ability for users to use the arrow keys to fix typos in prompts, but it does make the prompts visible until we have the time to get to highline 3.0+.

### DIFF
--- a/lib/manageiq/appliance_console/prompts.rb
+++ b/lib/manageiq/appliance_console/prompts.rb
@@ -212,7 +212,6 @@ module ApplianceConsole
 
     def just_ask(prompt, default = nil, validate = nil, error_text = nil, klass = nil)
       ask("Enter the #{prompt}: ", klass) do |q|
-        q.readline = true
         q.default = default.to_s if default
         q.validate = validate if validate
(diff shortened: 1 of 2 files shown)


lib/manageiq/appliance_console/prompts.rb
```diff
@@ -212,7 +212,6 @@ def ask_with_menu(prompt, options, default = nil, clear_screen_after = true)
 
     def just_ask(prompt, default = nil, validate = nil, error_text = nil, klass = nil)
       ask("Enter the #{prompt}: ", klass) do |q|
-        q.readline = true
```
jrafanie, line 215: Should we set it explicitly to false so we can easily update it "true" after we fix it?

bdunne: 🤷🏻‍♂️

jrafanie: I'm fine either way.

